PR #3, opened 2019-03-01
Issue/jperf 404 gather logs after node startup failed

wyrzyk:


CHANGELOG.md
```diff
@@ -23,6 +23,15 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-2.6.0...master
 
+### Added
+- Gather Thread Dumps during Jira startup.
+- Add support for `infrastructure:4.9`.
+
+### Fixed
+- Gather Jira logs even if the node doesn't start in time. Reslove [JPERF-404].
```
dagguh:
```suggestion
- Gather Jira logs even if the node doesn't start in time. Resolve [JPERF-404].
```

CHANGELOG.md
```diff
@@ -23,6 +23,15 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-2.6.0...master
 
+### Added
+- Gather Thread Dumps during Jira startup.
+- Add support for `infrastructure:4.9`.
```
dagguh:
```suggestion
- Add support for `infrastructure:4.9.0`.
```

dagguh: Which feature from 4.9.0 did we start supporting? We were compatible already.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneStoppedNode.kt
```diff
@@ -137,6 +144,7 @@ internal data class StandaloneStoppedNode(
             if (deadline < now()) {
                 throw Exception("$uri failed to get out of $statusQuo status within $timeout")
             }
+            threadDump.gather(ssh, "thread-dumps")
```
dagguh: The commit doesn't explain why we're gathering threads during startup, but my guess is it's in order to diagnose what took so long. If that's so, if a startup phase is slow and times out, we won't have the thread dumps from that phase, because we `gather` after `throw Exception("$uri failed to get out of $statusQuo status within $timeout")`

CHANGELOG.md
wyrzyk: Thread dumps

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneStoppedNode.kt
wyrzyk: We're not going to have the last thread dump from the phase.

CHANGELOG.md
dagguh: Ok, cool. We already cover that by "Gather Thread Dumps during Jira startup". Dependency bumps by themselves are not part of the contract. Only compatibility across major versions is a contract.
From the pragmatic perspective, it'd be nice to be able to drop support for 4.9.0 without breaking a contract. E.g. just like this PR drops support for 4.8.0.
It's fine to do that, because there should be no justifiable reason for consumers to avoid going from infrastructure:4.9.0 to infrastructure:4.10.0

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneStoppedNode.kt
dagguh: Do we want to avoid the last thread dump?

wyrzyk: No. The intention was to gather the dumps before the deadline, but IMO it doesn't matter and we can gather one extra log.

dagguh: Since it was the intention, I'm with with both ways

CHANGELOG.md
dagguh: - [ ] Drop the 4.9.0 support contract

wyrzyk:
> dagguh approved these changes on behalf of atlassian/jpt

jet lag? :smile:

dagguh:
> > dagguh approved these changes on behalf of atlassian/jpt
> 
> jet lag? 😄

I was going to sleep and I accidentally alt-tabbed to my inbox so I saw your force push 😛

PR #6, opened 2019-03-18
copy the db computer from source formula

kema2000:


src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
```diff
@@ -306,6 +306,7 @@ class DataCenterFormula private constructor(
             computer = formula.computer
```
dagguh: We require tests to pass for all PRs.
Because CI runtime requires some secrets (like AWS creds, Maven push creds and Git pushback creds), we do not automatically run code from PRs in our CI. Contributors have to set up their own CI. It's trivial in most repos, because they don't need any secrets (Maven and Git are for releases only), but `aws-infrastructure` and `aws-resources` naturally need AWS creds.
[Our CI takes 20-40 minutes to run](https://travis-ci.com/atlassian/aws-infrastructure/builds), so you can use temporary credentials for a single run. Or you can use static ones if you wish. Either way, we pass them the safest way, via [repository settings](https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings). We use the `DefaultAWSCredentialsProviderChain`, so that's why I simply set the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` env vars, like so:
image

I went ahead and cherry-picked your commit and reopened the PR from my fork, so you don't have to set it all up this time: https://github.com/atlassian/aws-infrastructure/pull/7

dagguh: Finally managed to do it in https://github.com/atlassian/aws-infrastructure/pull/9
It was tough. Note to self: go through the full contributor journey myself, no "admin shortcuts".
/cc @wyrzyk

kema2000: Thanks Maciej

PR #13, opened 2019-03-28
Mark 2.7.3 release

wyrzyk:


CHANGELOG.md
```diff
@@ -21,7 +21,10 @@ Adding a requirement of a major version of a dependency is breaking a contract.
 Dropping a requirement of a major version of a dependency is a new contract.
 
 ## [Unreleased]
-[Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-2.7.2...master
+[Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-2.7.3...master
+
+## [2.7.3] - 2019-03-28
```
dagguh: 2.7.3 is not released yet
I'd release it first and then mark it

wyrzyk: Why not release and mark in the same PR?

dagguh: Proofs:
* the [2.7.3 diff](https://github.com/atlassian/aws-infrastructure/compare/release-2.7.2...release-2.7.3) is empty
* there is no such Maven package: https://packages.atlassian.com/webapp/#/search/package/eyJzZWFyY2giOiJnYXZjIiwiYXJ0aWZhY3RJRCI6ImF3cy1pbmZyYXN0cnVjdHVyZSIsInZlcnNpb24iOiIyLjcuMyIsInNlbGVjdGVkUGFja2FnZVR5cGUiOnsiaWQiOiJnYXZjIiwiZGlzcGxheU5hbWUiOiJNYXZlbiBHQVZDIiwiaWNvbiI6InBvbSJ9fQ==

We cannot predict how the release will go. It might:
* fail due to:
  * external services
  * bad creds
* be released as a different SemVer than intended
* be postponed due to a freshly found critical bug

dagguh: > Why not release and mark in the same PR?

* this PR doesn't release
* this PR marks data, which can be true only after the actual release

wyrzyk: > We cannot predict how the release will go. It might:

If the release will fail, then there's not a big deal. We can revert the commit.

If we start to release before marking the change, there is a higher chance to make a mistake.
In `marking PR` there's a chance someone will catch the version isn't set correctly. And we can't
revert once released version.

Releasing and then marking is probably a clearer solution, as it doesn't introduce time when the changelog is not consistent (it's just outdated), but it's riskier.

wyrzyk:
> We currently do not have sufficient c5d.9xlarge capacity in the Availability Zone you requested (eu-west-1b). Our system will be working on provisioning additional capacity. You can currently get c5d.9xlarge capacity by not specifying an Availability Zone in your request or choosing eu-west-1a, eu-west-1c. (Service: AmazonEC2; Status Code: 500; Error Code: InsufficientInstanceCapacity; Request ID: 79b4f047-3ce0-4bb5-b823-f5d124b9d1dc)

CHANGELOG.md
dagguh: > We can revert the commit.

We often don't, due to laziness. The release dates were sometimes postponed, but we didn't go back and fix the dates.
With the release-then-mark approach we do not have to revert.

> In marking PR there's a chance someone will catch the version isn't set correctly.
Do you mean the release tag? That can be done by the releaser without the PR. And PRs do not give good visibility for tags. And if we switch from tag-tracking to source-tracking, then we don't have to check it before release anyway, because the version was already accepted in previously merged PRs.

I "requested changes" to force this discussion. I'm happy with the arguments presented on both sides. I'll approve the changes now and let you decide the approach.

wyrzyk: > Do you mean the release tag? That can be done by the releaser without the PR. 

Yes. But PR gives a second pair of eyes.

> And PRs do not give good visibility for tags. And if we switch from tag-tracking to source-tracking, then we don't have to check it before release anyway, because the version was already accepted in previously merged PRs.

I'm for source-tracking from the beginning :)

dagguh: > Yes. But PR gives a second pair of eyes.

TBH I never look at the tags when I review the changelog PRs. Because it's too many clicks and because the releaser should check it anyway.
But I do always look at the tags before I actually release.

PR #15, opened 2019-04-08
JPERF-451: Give more control over `VirtualUserOptions`

dagguh:


src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/Infrastructure.kt
```diff
@@ -73,6 +80,29 @@ class Infrastructure<out T : VirtualUsers>(
...
+
+    override fun target(
+        jira: URI
+    ): VirtualUserOptions = VirtualUserOptions(
+        target = VirtualUserTarget(
+            webApplication = jira,
+            userName = "admin",
+            password = "admin"
+        ),
+        behavior = behavior
```
dagguh: @sebapawlak, @wyrzyk you were right, the arbitrary behavior/target split was not the proper long-term solution.
We were trying to let aws-infra inject the Jira URI into VUO. It turns out we can ask for this directly 🤦‍♂️

dagguh:
CI failed due to:
> We currently do not have sufficient c5d.9xlarge capacity in the Availability Zone you requested (eu-west-1b)
@sebapawlak did you see a similar symptom on our dev account as well?

sebapawlak:
> > We currently do not have sufficient c5d.9xlarge capacity in the Availability Zone you requested (eu-west-1b)

dagguh:
Close/reopen seems to help in cases when you open a PR before enabling CircleCI. If it is already enabled, then you can rerun from CircleCI itself. I just triggered a rebuild.

dagguh:
The capacity problem happened again in the rerun: https://circleci.com/gh/dagguh/aws-infrastructure/9

PR #17, opened 2019-04-16
JPERF-454 Fix c5d.9xlarge capacity problem

wyrzyk:
We had the same problem running internal tests.
Switching to use only "eu-west-1a", "eu-west-1c" resolved the issue.

### DIFF
--- a/src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/IntegrationTestRuntime.kt
+++ b/src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/IntegrationTestRuntime.kt
@@ -1,10 +1,10 @@
 package com.atlassian.performance.tools.awsinfrastructure
 
-import com.amazonaws.auth.DefaultAWSCredentialsProviderChain
 import com.amazonaws.regions.Regions
 import com.atlassian.performance.tools.aws.api.Aws
 import com.atlassian.performance.tools.workspace.api.RootWorkspace
 import org.apache.logging.log4j.core.config.ConfigurationFactory
+import java.util.function.Predicate
 
 object IntegrationTestRuntime {
     val taskWorkspace = RootWorkspace().currentTask
@@ -12,9 +12,9 @@ object IntegrationTestRuntime {
 
     init {
         ConfigurationFactory.setConfigurationFactory(LogConfigurationFactory(taskWorkspace))
-        aws = Aws(
-            region = Regions.EU_WEST_1,
-            credentialsProvider = DefaultAWSCredentialsProviderChain()
-        )
+        aws = Aws.Builder(Regions.EU_WEST_1)
+            .availabilityZoneFilter(Predicate { it.zoneName in listOf("eu-west-1a", "eu-west-1c") })
+            .regionsWithHousekeeping(listOf(Regions.EU_WEST_1))
+            .build()
     }
 }
--- a/gradle/dependency-locks/apiDependenciesMetadata.lockfile
+++ b/gradle/dependency-locks/apiDependenciesMetadata.lockfile
@@ -7,11 +7,12 @@ com.amazonaws:aws-java-sdk-ec2:1.11.424
 com.amazonaws:aws-java-sdk-elasticloadbalancing:1.11.424
 com.amazonaws:aws-java-sdk-iam:1.11.424
 com.amazonaws:aws-java-sdk-kms:1.11.424
+com.amazonaws:aws-java-sdk-rds:1.11.424
 com.amazonaws:aws-java-sdk-s3:1.11.424
 com.amazonaws:aws-java-sdk-sts:1.11.424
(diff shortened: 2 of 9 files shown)


src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/IntegrationTestRuntime.kt
```diff
@@ -1,20 +1,20 @@
...
     val aws: Aws
 
     init {
         ConfigurationFactory.setConfigurationFactory(LogConfigurationFactory(taskWorkspace))
-        aws = Aws(
-            region = Regions.EU_WEST_1,
-            credentialsProvider = DefaultAWSCredentialsProviderChain()
-        )
+        aws = Aws.Builder(Regions.EU_WEST_1)
+            .availabilityZoneFilter(Predicate { it.zoneName in listOf("eu-west-1a", "eu-west-1c") })
```
dagguh: Alternatively, we can blacklist `eu-west-1b` instead of whitelisting A and C. If they add a D AZ, we'd pick it up transparently.

src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/IntegrationTestRuntime.kt
```diff
@@ -1,20 +1,20 @@
...
+            .regionsWithHousekeeping(listOf(Regions.EU_WEST_1))
```
dagguh: Where's the proof of houskeeping in `EU_WEST_1`? In BB Pipelines we had a scheduled build, but now we do not have it.

src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/IntegrationTestRuntime.kt
wyrzyk: > Alternatively, we can blacklist `eu-west-1b` instead of whitelisting A and C. If they add a D AZ, we'd pick it up transparently.

That's why I didn't want to blacklist. If they add a D AZ, and it will have no capacity for our instances, then our builds will start to fail again.

dagguh: OTOH if there's more capacity in D, we won't use it

dagguh: Also, if you change the region, you'll have to change the whitelist too. But a blacklist would be immune to this change.

src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/IntegrationTestRuntime.kt
wyrzyk: We have housekeeping set in bamboo.

dagguh: That's a remote contract. If the internal repo migrates to another region, this CI will start leaking AWS resources. Also, it's extra risk for contributors which run on another AWS account, e.g. 3rd party or non-Jira Atlassians.

src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/IntegrationTestRuntime.kt
wyrzyk: It's the same problem that we have with other dependencies. We can have new features and bug fixes in the latest versions, but we also can get a new bug. We choose to lock dependencies for our builds to make them repeatable and not to disturb our work. Each time we bump dependencies we do it consciously. IMO we should stick to this strategy also for infrastructure dependencies.

dagguh: Fewer moving parts, fair enough 👍

src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/IntegrationTestRuntime.kt
wyrzyk: This PR doesn't introduce the risk.

dagguh: It circumvents the deprecation warning that was there before this PR. Deprecation of old "unsafe" API was a mechanism to spread the warning until we really fix it. Contributors are impacted either way, so I guess it's not that much worse.

dagguh: Raised https://ecosystem.atlassian.net/browse/JPERF-458

PR #19, opened 2019-04-18
Allow jira-config.properties file to become part of the dataset.

bornatow:
The autoexport property will be added only if it does not already exist in the file or the file does not exist at all. In other words, the jira-config.properties file that has already been restored with the dataset will remain unchanged or will get the autoexport=false property appended to it.

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneNodeFormula.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneNodeFormula.kt
@@ -52,7 +52,8 @@ internal class StandaloneNodeFormula(
                 jiraIp = ssh.host.ipAddress
             )
             connection.execute("echo jira.home=`realpath $jiraHome` > $unpackedProduct/atlassian-jira/WEB-INF/classes/jira-application.properties")
-            connection.execute("echo jira.autoexport=false > $jiraHome/jira-config.properties")
+            connection.execute("touch $jiraHome/jira-config.properties")
+            connection.execute("grep -qxF \"jira.autoexport=false\" $jiraHome/jira-config.properties || echo -e \"\\njira.autoexport=false\" >> $jiraHome/jira-config.properties")
             downloadMysqlConnector(
                 "https://dev.mysql.com/get/Downloads/Connector-J/mysql-connector-java-5.1.40.tar.gz",
                 connection


wyrzyk:
I've created an issue (https://ecosystem.atlassian.net/browse/JPERF-463). Please add the issue key to the commit message and update CHANGELOG.md file.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneNodeFormula.kt
```diff
@@ -52,7 +52,8 @@ internal class StandaloneNodeFormula(
                 jiraIp = ssh.host.ipAddress
             )
             connection.execute("echo jira.home=`realpath $jiraHome` > $unpackedProduct/atlassian-jira/WEB-INF/classes/jira-application.properties")
-            connection.execute("echo jira.autoexport=false > $jiraHome/jira-config.properties")
+            connection.execute("touch $jiraHome/jira-config.properties")
+            connection.execute("grep -qxF \"jira.autoexport=false\" $jiraHome/jira-config.properties || echo -e \"\\njira.autoexport=false\" >> $jiraHome/jira-config.properties")
```
wyrzyk: It looks error prone. What if we add `jira.autoexport = false`. It will still work, but will not be detected by grep.
It's still better than before, so I would not block the review 👍 


We may also not need the property. I think it doesn't work like we expected (https://ecosystem.atlassian.net/browse/JPERF-344).

wyrzyk: And AFAIK all our datasets already have the property set :)

bornatow: I think we could improve the grep, but I am going away till the end of Apr, so could do that in 2 weeks time. And yes, all our datasets have that already, my problem was that the jira-config.properties file was always overwritten, so I changed this to be less invasive. But if we can get rid of this altogether through the issue you linked, then I would be just as happy.

wyrzyk: 👍  Just add the issue key and changelog and we can merge. (Remember to rebase!)

dagguh: If this fixes JPERF-344, I wonder if the integration tests will become faster here. I hope we see it in the CI results.

wyrzyk: > If this fixes JPERF-344, I wonder if the integration tests will become faster here. I hope we see it in the CI results.

It doesn't fix JPERF-344. AFAIK we see the export even if the flag is set.

pczuj (approved):

PR #21, opened 2019-05-22
RUM-2699: Change load balancer to apache proxy

marcinsasinowski:


src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/loadbalancer/ApacheEc2LoadBalancerFormula.kt
```diff
@@ -82,25 +86,48 @@ internal class ApacheEc2LoadBalancerFormula : LoadBalancerFormula {
...
-        )
+
     }
+
+    private val APACHE_CONFIG_PATH = "/etc/apache2/sites-enabled/000-default.conf"
+
+    fun provision() {
+        ssh.newConnection().use { connection ->
+            connection.safeExecute("sudo apt-get update", Duration.ofMinutes(3))
+            connection.safeExecute("sudo apt-get --assume-yes install apache2", Duration.ofMinutes(1))
```
dagguh: We can reuse `Ubuntu().install` for a few improvements, e.g. error checking, retries, race-condition control.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/loadbalancer/ApacheEc2LoadBalancerFormula.kt
```diff
@@ -82,25 +86,48 @@ internal class ApacheEc2LoadBalancerFormula : LoadBalancerFormula {
...
+            connection.safeExecute("sudo rm $APACHE_CONFIG_PATH")
+            connection.safeExecute("sudo touch $APACHE_CONFIG_PATH")
```
dagguh: `safeExecute` will ignore the exit code. If you want to check it out-of-the-box, then you can switch to `execute`

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/loadbalancer/ApacheEc2LoadBalancerFormula.kt
```diff
@@ -82,25 +86,48 @@ internal class ApacheEc2LoadBalancerFormula : LoadBalancerFormula {
...
-    init {
-        throw Exception("Wait until the `infrastructure` module ships `ApacheLoadBalancer`")
-    }
-
     override val uri: URI = URI("http://${ssh.host.ipAddress}:$httpPort/")
 
     override fun waitUntilHealthy(
         timeout: Duration
     ) {
-        throw Exception(
```
dagguh: This impl is empty. Is the `uri` available immediately?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
```diff
@@ -286,7 +286,7 @@ class DataCenterFormula private constructor(
         )
 
         private var configs: List<JiraNodeConfig> = (1..2).map { JiraNodeConfig.Builder().name("jira-node-$it").build() }
-        private var loadBalancerFormula: LoadBalancerFormula = ElasticLoadBalancerFormula()
+        private var loadBalancerFormula: LoadBalancerFormula = ApacheEc2LoadBalancerFormula()
```
dagguh: We don't need to change the default. We can offer `ApacheEc2LoadBalancerFormula` and the users can choose whether they want to use it or `ElasticLoadBalancerFormula`

src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormulaIT.kt
```diff
@@ -14,13 +14,19 @@ import com.atlassian.performance.tools.workspace.api.TaskWorkspace
 import com.atlassian.performance.tools.workspace.api.TestWorkspace
 import com.google.common.util.concurrent.ThreadFactoryBuilder
 import org.apache.logging.log4j.CloseableThreadContext
-import org.assertj.core.api.Assertions.assertThat
```
dagguh: We usually migrate away from Hamcrest and towards AssertJ.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/loadbalancer/ApacheEc2LoadBalancerFormula.kt
```diff
@@ -35,16 +36,19 @@ internal class ApacheEc2LoadBalancerFormula : LoadBalancerFormula {
                 launch
                     .withSecurityGroupIds(httpAccess.groupId)
                     .withSubnetId(subnet.subnetId)
-                    .withInstanceType(InstanceType.M1Large)
+                    .withInstanceType(InstanceType.M5Large)
             }
         )
+        key.file.facilitateSsh(ssh.host.ipAddress)
+        val loadBalancer = HaProxyLoadBalancer(
```
dagguh: If we want users to consume `ApacheEc2LoadBalancerFormula`, it should be moved to the `api` subpackage and we should mention we `### Added` new API in the `CHANGELOG.md`

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
sebapawlak: The problem is that the current default behaves in an ill way. It does not distribute the load equally.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/loadbalancer/ApacheEc2LoadBalancerFormula.kt
sebapawlak: We definitely want this available for users to consume (including internal Jira perf. tests). Let's move to api and adjust the `CHANGELOG.md` accordingly.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
dagguh: 1. It would be nice to see that problem in a red-test-first way.
2. We can still offer the bug fix without changing the default.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/loadbalancer/ApacheEc2LoadBalancerFormula.kt
marcinsasinowski: It will available immediately because provision method is restarting apache2 service at the end and the call is blocking.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/loadbalancer/ApacheEc2LoadBalancerFormula.kt
marcinsasinowski: Done.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/loadbalancer/ApacheEc2LoadBalancerFormula.kt
marcinsasinowski: Done.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
marcinsasinowski: Regarding 1. it will be hard to do because the test is working based on Cookie set by load balancer (won't be working with other load balancers)

dagguh: Makes sense.
We measure load balancer fairness in a LB-agnostic way in VUs, who identify the node from the footer in the HTML UI.
We could integrate VUs or some other cookie-aware HTML-sniffing, but feel free to skip that. We did not test fairness before, so I don't want to impose the dev cost onto you.

src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormulaIT.kt
marcinsasinowski: Done.

marcinsasinowski:
Declined in favour of https://github.com/atlassian/aws-infrastructure/pull/22

PR #22, opened 2019-05-27
JPERF-486: Change load balancer to apache proxy

marcinsasinowski:


src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/ApacheEc2LoadBalancerFormula.kt
```diff
@@ -1,12 +1,13 @@
-package com.atlassian.performance.tools.awsinfrastructure.loadbalancer
+package com.atlassian.performance.tools.awsinfrastructure.api.loadbalancer
```
dagguh: We moved it to `api`, but it's still `internal`

PR #24, opened 2019-06-19
Add parameter for specifying volume size

swidnow2:
Extracted @kema2000 changes from postgresDbSupport branch.
This is to get out from snapshot versions in jira-hardware-exploration.
[Related ticket](https://ecosystem.atlassian.net/browse/JPERF-26)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-2.9.0...master
 
+### Added
+- Add the parameter for specifying volume size. Resolves [JPERF-26]
+
+[JPERF-26]: https://ecosystem.atlassian.net/browse/JPERF-26
+
 ## [2.9.0] - 2019-05-28
 [2.9.0]: https://github.com/atlassian/aws-infrastructure/compare/release-2.8.0...release-2.9.0
 
--- a/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/hardware/Volume.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/hardware/Volume.kt
@@ -1,0 +1,3 @@
+package com.atlassian.performance.tools.awsinfrastructure.api.hardware
+
+class Volume(val size: Int)
--- a/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
@@ -11,6 +11,7 @@ import com.atlassian.performance.tools.awsinfrastructure.api.RemoteLocation
 import com.atlassian.performance.tools.awsinfrastructure.api.hardware.C4EightExtraLargeElastic
 import com.atlassian.performance.tools.awsinfrastructure.api.hardware.Computer
 import com.atlassian.performance.tools.awsinfrastructure.api.hardware.M4ExtraLargeElastic
+import com.atlassian.performance.tools.awsinfrastructure.api.hardware.Volume
 import com.atlassian.performance.tools.awsinfrastructure.api.loadbalancer.ApacheEc2LoadBalancerFormula
 import com.atlassian.performance.tools.awsinfrastructure.api.loadbalancer.LoadBalancerFormula
 import com.atlassian.performance.tools.awsinfrastructure.jira.DataCenterNodeFormula
@@ -46,9 +47,11 @@ class DataCenterFormula private constructor(
     private val jiraHomeSource: JiraHomeSource,
     private val database: Database,
     private val computer: Computer,
+    private val jiraVolume: Volume,
     private val stackCreationTimeout: Duration,
     private val overriddenNetwork: Network? = null,
-    private val databaseComputer: Computer
+    private val databaseComputer: Computer,
+    private val databaseVolume: Volume
 ) : JiraFormula {
     private val logger: Logger = LogManager.getLogger(this::class.java)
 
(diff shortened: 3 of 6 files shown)


CHANGELOG.md
```diff
@@ -21,7 +21,13 @@ Adding a requirement of a major version of a dependency is breaking a contract.
...
 
 ## [Unreleased]
-[Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-2.9.0...master
+[Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-2.10.0...master
+
+## [2.10.0] - 2019-06-19
+[2.10.0]: https://github.com/atlassian/aws-infrastructure/compare/release-2.9.0...release-2.10.0
+
+### Added
+- Add the parameter for specifying volume size.
```
wyrzyk: Can you point to the issue?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/hardware/Computer.kt
```diff
@@ -8,4 +8,8 @@ abstract class Computer internal constructor() {
     internal abstract val instanceType: InstanceType
 
     internal abstract fun setUp(ssh: SshConnection)
+
+    var volumeSize = 100
+
+    fun withVolumeSize(volumeSize : Int) = apply { this.volumeSize = volumeSize }
```
wyrzyk: I'm not sure if it's not too specific. AFAIR volume has at least 3 parameters (size, IOPS, type). 
```
withVolume(volume : Volume)`

class Volume(val size);
``

It could be more open to extending in the future.

File: CHANGELOG.md (hunk 1 above)
Comment (swidnow2) [reply CHANGELOG.md]: Done

File: src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/hardware/Volume.kt [hunk 3]
```diff
@@ -0,0 +1,7 @@
+package com.atlassian.performance.tools.awsinfrastructure.api.hardware
+
+class Volume {
```
wyrzyk: Can you make it immutable?

src/main/resources/aws/single-node.yaml
```diff
@@ -62,7 +66,7 @@ Resources:
         -
           DeviceName: /dev/sda1
           Ebs:
-            VolumeSize: 200
```
wyrzyk: It was 200 before, and the commit changes it to 100. Volume size affects IOPS. The instance can be slower for a smaller drive. Can you explain the change?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/hardware/Computer.kt
```diff
@@ -8,4 +8,8 @@ abstract class Computer internal constructor() {
     internal abstract val instanceType: InstanceType
 
     internal abstract fun setUp(ssh: SshConnection)
+
+    var volume = Volume()
```
wyrzyk: Does it need to be public?

src/main/resources/aws/single-node.yaml
dagguh: The change adds a common abstraction over both DC and Server volumes. It didn't exist before, so we're translating inconsistency into consistency. This change is fundamentally incompatible.

### Parametrize elsewhere
We can avoid this forced consistency by moving the parametrization out of `Computer` and into `DataCenterFormula.Builder` and `StandaloneFormula.Builder`.
This does look like it belongs to `Computer`, but we can clean up that design during the next major version.

### Override
We override the `Volume` in `StandaloneFormula` to 200 GB.
However, it will ignore the user-supplied value.
We can override only if the user-supplied value is the default, ie. 100 GB. But what if the user wanted to actually use 100 GB explicitly? We can use a dirty trick. We make the default a very specific value, e.g 99.71 GB and detect that and hope that no user would specify this special value.

CHANGELOG.md
```diff
@@ -21,7 +21,13 @@ Adding a requirement of a major version of a dependency is breaking a contract.
 Dropping a requirement of a major version of a dependency is a new contract.
 
 ## [Unreleased]
-[Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-2.9.0...master
+[Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-2.10.0...master
+
+## [2.10.0] - 2019-06-19
```
dagguh: You don't have to add the release markers. We'll do it after the release.

dagguh:
@swidnow2, you are cleaning up after me and @kema2000. I appreciate it very much.
If you find there's too many obstacles, let me know, I'll talk with my team to clean up after ourselves. We have the context, we remember (more or less) our past intentions, etc.

src/main/resources/aws/single-node.yaml
dagguh: There's another option, we can accept the forced consistency. Instead of demoting 200 GB to 100 GB, we can bump 100 GB DC to 200 GB DC

100 GB of GP2 EBS costs 1,5 US cents an hour. For a 1-hour test on a 2-node cluster that's 3 cents. Sounds acceptable, especially since there's a workaround.
There might be another consequence in form of longer provisioning queues, but I have not seen EBS as the scarce resource yet.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
```diff
@@ -46,9 +47,11 @@ class DataCenterFormula private constructor(
     private val jiraHomeSource: JiraHomeSource,
     private val database: Database,
     private val computer: Computer,
+    private val jiraVolume: Volume = Volume(100),
     private val stackCreationTimeout: Duration,
     private val overriddenNetwork: Network? = null,
-    private val databaseComputer: Computer
+    private val databaseComputer: Computer,
+    private val databaseVolume: Volume = Volume(100)
```
dagguh: The constructor is private, so we can add parameters without defaults.
And these defaults are not necessary. They're already expressed via the builder.

In general, defaults are a smell. However, this is `private`, so the smell doesn't leak out to the API. Therefore I do not require this to be fixed.

PR #29, opened 2019-06-27
JPERF-506: Fix apache load balancer problem with Jira

marcinsasinowski:


src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/ApacheEc2LoadBalancerFormula.kt
```diff
@@ -13,7 +14,9 @@ import org.apache.logging.log4j.Logger
 import java.net.URI
 import java.time.Duration
 
-internal class ApacheEc2LoadBalancerFormula : LoadBalancerFormula {
+class ApacheEc2LoadBalancerFormula : LoadBalancerFormula {
```
sebapawlak: Theoretically we are adding new API here so it looks like this should belong to `### Added` section and a minor bump.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/ElasticLoadBalancerFormula.kt
```diff
@@ -50,4 +51,9 @@ class ElasticLoadBalancerFormula : LoadBalancerFormula {
             resource = stack
         )
     }
+
+    override fun updateJiraConfiguration(
+        ssh: Ssh,
+        unpackedProduct: String) {
```
sebapawlak: Formatting here. `) {` should be in the next line.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/LoadBalancerFormula.kt
```diff
@@ -19,6 +20,10 @@ interface LoadBalancerFormula {
         key: SshKey,
         aws: Aws
     ): ProvisionedLoadBalancer
+
+    fun updateJiraConfiguration(
+        ssh: Ssh,
+        unpackedProduct: String)
```
sebapawlak: Formatting.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/LoadBalancerFormula.kt
```diff
@@ -19,6 +20,10 @@ interface LoadBalancerFormula {
         key: SshKey,
         aws: Aws
     ): ProvisionedLoadBalancer
+
+    fun updateJiraConfiguration(
```
sebapawlak: This an interface in the API package which means this is an SPI. Adding new method without default implementation means API-breaking change. I suggest adding a default no-op implementation.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/ApacheEc2LoadBalancerFormula.kt
```diff
@@ -56,6 +60,12 @@ internal class ApacheEc2LoadBalancerFormula : LoadBalancerFormula {
         )
     }
 
+    override fun updateJiraConfiguration(ssh: Ssh, unpackedProduct: String) {
+        Sed().replace(ssh.newConnection(), "bindOnInit=\"false\"" ,
```
wyrzyk: connection leak.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/ApacheEc2LoadBalancerFormula.kt
marcinsasinowski: Done.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/ElasticLoadBalancerFormula.kt
marcinsasinowski: Done.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/LoadBalancerFormula.kt
marcinsasinowski: Done.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/LoadBalancerFormula.kt
marcinsasinowski: Done.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/ApacheEc2LoadBalancerFormula.kt
marcinsasinowski: Done.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/ApacheEc2LoadBalancerFormula.kt
```diff
@@ -56,6 +60,14 @@ internal class ApacheEc2LoadBalancerFormula : LoadBalancerFormula {
         )
     }
 
+    override fun updateJiraConfiguration(ssh: Ssh, unpackedProduct: String) {
+        ssh.newConnection().use { shell ->
+            Sed().replace(shell, "bindOnInit=\"false\"",
```
wyrzyk: IMO we use slightly different code style in the project:

```
Sed().replace(
    shell,
    "bindOnInit=\"false\"",
    "bindOnInit=\"false\" scheme=\"http\" proxyName=\"${uri?.host}\" proxyPort=\"80\"",
    "$unpackedProduct/conf/server.xml"
)
```

marcinsasinowski: Done.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StoppedNode.kt
```diff
@@ -6,5 +6,6 @@ import com.atlassian.performance.tools.ssh.api.Ssh
 internal interface StoppedNode {
     val jiraHome: String
     val ssh: Ssh
+    val unpackedProduct: String
```
wyrzyk: All the configurations until now were done in `StoppedNode`. Is it possible to keep that convention? I have three ideas that could work here:
1. Can a Node accept a list of  `fun updateConfiguration(ssh, jiraHome, unpackedProduct)` functions and run it after the internal configuration?
2. Can we pass LB future to a stopped node? Does it make sense? 
3. Can we expose  `fun updateConfiguration(ssh, jiraHome, unpackedProduct)` instead of unpackedProduct?

@marcinsasinowski, @dagguh WDYT?

marcinsasinowski: I was thinking about options you presented and I will choose the 1st option (still might be premature, but let's do it :) )

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/ApacheEc2LoadBalancerFormula.kt
```diff
@@ -56,6 +60,16 @@ internal class ApacheEc2LoadBalancerFormula : LoadBalancerFormula {
         )
     }
 
+    override fun updateJiraConfiguration(ssh: Ssh, unpackedProduct: String) {
+        ssh.newConnection().use { shell ->
+            Sed().replace(
+                shell,
+                "bindOnInit=\"false\"",
+                "bindOnInit=\"false\" scheme=\"http\" proxyName=\"${uri?.host}\" proxyPort=\"80\"",
```
wyrzyk: IMO it's better to fail on null here instead of allowing to silently break the configuration.

dagguh: Currently it would produce `proxyName="null"`

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/ApacheEc2LoadBalancerFormula.kt
```diff
@@ -13,7 +14,9 @@ import org.apache.logging.log4j.Logger
 import java.net.URI
 import java.time.Duration
 
-internal class ApacheEc2LoadBalancerFormula : LoadBalancerFormula {
+class ApacheEc2LoadBalancerFormula : LoadBalancerFormula {
+
+    private var uri: URI? = null
```
dagguh: This mutability is a smell. There's a hidden contract between `provision` and `updateJiraConfiguration`. One must be called after another.
And it is bad for thread-safety if you reuse the same LB formula for two cohorts.

Perhaps `updateJiraConfiguration` should be part of `LoadBalancer` API.

marcinsasinowski: I'll add `updateJiraConfiguration` to part of `ProvisionedLoadBalancer`.

dagguh: Could the `LoadBalancer` have it? This way we could empower all SSH-based use cases, not only AWS-SSH ones. We could move `HaProxyLoadBalancer` to `infrastructure`, which has a wider audience and a more lightweight dev loop.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/LoadBalancerFormula.kt
```diff
@@ -23,5 +24,13 @@ interface LoadBalancerFormula {
 
 class ProvisionedLoadBalancer(
     val loadBalancer: LoadBalancer,
-    val resource: Resource
-)
\ No newline at end of file
+    val resource: Resource,
+    val updateJiraConfigurationFunc: (ssh: Ssh, unpackedProduct: String) -> Unit
```
marcinsasinowski: This should be in second constructor in order not to break API.

marcinsasinowski: Fix.ed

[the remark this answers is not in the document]
dagguh: Ok, it's here to be overridden by some instantiators of `ProvisionedLoadBalancer` and it's better than making the class `open` 👍

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/ApacheEc2LoadBalancerFormula.kt
```diff
@@ -86,7 +100,7 @@ internal class ApacheEc2LoadBalancerFormula : LoadBalancerFormula {
     }
 }
 
-private class HaProxyLoadBalancer(
+private class ApacheProxyLoadBalancer(
     private val nodes: List<URI>,
     private val ssh: Ssh,
     private val httpPort: Int
```
wyrzyk: warning: ```Constructor parameter is never used as a property```

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/ElasticLoadBalancerFormula.kt
```diff
@@ -10,6 +10,7 @@ import com.atlassian.performance.tools.aws.api.SshKey
 import com.atlassian.performance.tools.aws.api.StackFormula
 import com.atlassian.performance.tools.awsinfrastructure.loadbalancer.ElasticLoadBalancer
 import com.atlassian.performance.tools.io.api.readResourceText
+import com.atlassian.performance.tools.ssh.api.Ssh
```
wyrzyk: Unused import.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/ApacheEc2LoadBalancerFormula.kt
marcinsasinowski: Done.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/LoadBalancerFormula.kt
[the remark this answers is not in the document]
marcinsasinowski: I've changed that.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/ApacheEc2LoadBalancerFormula.kt
marcinsasinowski: Done.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/ElasticLoadBalancerFormula.kt
marcinsasinowski: Done.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
```diff
@@ -251,9 +252,13 @@ class DataCenterFormula private constructor(
 
         val databaseDataLocation = setupDatabase.get()
 
+        val updateJiraConfiguration =
+            if (loadBalancer is ApacheProxyLoadBalancer) listOf(loadBalancer::updateJiraConfiguration) else emptyList()
```
dagguh: ;)

wyrzyk: @dagguh We can think about an abstraction if more cases appear in the if statement.

dagguh: Yeah, hopefully [JPERF-273](https://ecosystem.atlassian.net/browse/JPERF-273) will unblock all of that

PR #30, opened 2019-06-27
JPERF-509: Add `AwsDatasetModification`

mcmurphy2000:


src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
```diff
@@ -92,7 +74,40 @@ internal class AwsDataset(
         logger.info("AWS resources released")
     }
 
+    class Builder(
+        private val aws: Aws,
+        private val dataset: Dataset
+    ) {
+        private var workspace: TestWorkspace = RootWorkspace().currentTask.isolateTest(javaClass.simpleName) // TestWorkspace(Files.createTempDirectory("dataset-modification-workspace"))
```
mcmurphy2000: Not sure what is the proper way to create temp workspace

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
```diff
@@ -1,77 +1,59 @@
...
+): Function<Consumer<Infrastructure<*>>, Dataset> {
 
-    fun modify(
-        aws: Aws,
-        workspace: TestWorkspace,
-        newDatasetName: String,
-        modification: (Infrastructure<*>) -> Unit
-    ): Dataset {
-        val provisionedInfrastructure = provision(aws, workspace)
+    override fun apply(modification: Consumer<Infrastructure<*>>): Dataset {
```
drspartez: Having Consumer here means that we will simply  throw exceptions from transformer if something was wrong and we should not save the new data set. Is this expected behaviour ?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
```diff
@@ -1,77 +1,59 @@
...
 import java.time.Duration
+import java.util.*
 import java.util.concurrent.TimeUnit
+import java.util.function.Consumer
+import java.util.function.Function
 
-internal class AwsDataset(
-    private val dataset: Dataset
-) {
+class AwsDatasetModification private constructor(
```
drspartez: I do not see any cleanup code for extra plugins - is there decision to do it in transformer ?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
mcmurphy2000: Hmm... Maybe we should use Predicate instead of Consumer and only save dataset if it returns true?
This has also just given me an idea to surround `apply` with `try...finally` to make sure we release resource properly

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
mcmurphy2000: @dagguh has suggested to do cleaning inside the Consumer

drspartez: Ok,  should the consumer be aware of jira is still running while it needs to do some cleanup ?

dagguh: There should be hardcoded cleanup for license and plugins.
If the modification itself requires additional cleanup, then it has to be a part of the modification.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
dagguh: +1 for try-finally
-1 for `Predicate`, I don't see the added value, let the `Exception`s fly

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
dagguh: Good call on the jira state. Some things might need cleanup after it is stopped. We can have `onlineModification` and `offlineModification`

mcmurphy2000: > Standard/default cleanup should be hardcoded in. It should include licenses and plugins.
How can we hardcode cleaning-up plugins if we don't know which plugins might have been provided via `InfrastructureFormula<*>`?

> We can have onlineModification and offlineModification
Should `offlineModification` be applied BEFORE Jira is started? Or should we stop Jira at some point and then apply this `offlineModification`?

dagguh: Nuke all the plugins.

TBH I don't know if we need pre-start or post-stop mods. Perhaps we're adding API unnecessarily? I don't know of any use-cases for those yet. And we're using the builder pattern, so it's trivial to add it in later when such use cases pop up.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
```diff
@@ -92,7 +74,40 @@ internal class AwsDataset(
...
+        fun build() = AwsDatasetModification(
+            aws = aws,
+            dataset = dataset,
+            workspace = workspace,
+            newDatasetName = newDatasetName,
+            formula = formula
+        )
+    }
+
     companion object {
```
wyrzyk: The companion object could be private.

src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/AwsDatasetModificationIT.kt
```diff
@@ -4,38 +4,43 @@ import com.amazonaws.regions.Regions
...
 import com.atlassian.performance.tools.infrastructure.api.dataset.Dataset
 import com.atlassian.performance.tools.ssh.api.Ssh
 import org.junit.Test
 import java.net.URI
 import java.time.Duration.ofMinutes
-import java.util.*
+import java.util.function.Consumer
 
-class AwsDatasetIT {
+class AwsDatasetModificationIT {
```
wyrzyk: The test's package could follow a package of the tested class.

wyrzyk:
Please update CHANGELOG.md

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
```diff
@@ -74,6 +80,21 @@ class AwsDatasetModification private constructor(
         logger.info("AWS resources released")
     }
 
+    private fun cleanUp(infrastructure: Infrastructure<*>) {
+        val jiraHome = infrastructure.jira.jiraHome
+        val dirsToPurge = listOf(
```
sebapawlak: should we purge import and export directories as well just in case? @wyrzyk what do you think?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
```diff
@@ -1,77 +1,59 @@
...
-internal class AwsDataset(
-    private val dataset: Dataset
-) {
+class AwsDatasetModification private constructor(
+    private val aws: Aws,
+    private val dataset: Dataset,
+    private val workspace: TestWorkspace,
+    private val newDatasetName: String,
+    private val formula: InfrastructureFormula<*>
+): Function<Consumer<Infrastructure<*>>, Dataset> {
```
mcmurphy2000: @wyrzyk @dagguh @drspartez do you think we should use normal class instead of implementing `Function<Consumer<Infrastructure<*>>, Dataset>` interface here? So if we want to add another public metho sin the future we won't have to break the API?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
wyrzyk: 👍

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
mcmurphy2000: @wyrzyk @dagguh @drspartez do you think `modification: Consumer<Infrastructure<*>>` should be a parameter to the constructor instead (with default no-op implementation in Builder)?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
mcmurphy2000: @wyrzyk @dagguh could you suggest a proper way to create a default temp workspace pls?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
wyrzyk: WDYM? Wy adding a method to ADM would break an API?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
wyrzyk: 🤷‍♂

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
wyrzyk: 1. No-op seems not to be a good default. 
2. We started to avoid defaults in a constructor in favour of builders. It'll be hard/impossible to consume the default value from java.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
mcmurphy2000: Sorry, I meant adding new parameters. But that would break API anyway so nevermind

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
drspartez: We don't know now - that's fair. But we may suspect or imagine such case. I think that replacing consumer with similar, but  homemade interface would provide us more flexibility in the future  
(we can simply add new method for some offline mod instead replacing interface)

wyrzyk: @drspartez current implementation doesn't block the flexibility in the future. There's no need to change the interface, we can add a new one.

drspartez: That is why i'd liek to avoid that. I do prefer keeping contract in one place. With multiple interfaces we make complexity. This is good practice (adding new interfaces for existing impl) to do change in legacy systems. Here we could look little bit far beyond. This cost nothing.

wyrzyk: You're right. It's a little change and gives us more room in the future.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
```diff
@@ -0,0 +1,136 @@
...
+import java.util.concurrent.TimeUnit
+import java.util.function.Function
+
+class AwsDatasetModification private constructor(
+    private val aws: Aws,
+    private val dataset: Dataset,
+    private val workspace: TestWorkspace,
+    private val newDatasetName: String,
+    private val formula: InfrastructureFormula<*>
+): Function<DatasetTransformation, Dataset> {
```
dagguh: Why did we translate `Consumer<Infrastructure<*>` into `DatasetTransformation`?

src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModificationIT.kt
```diff
@@ -1,30 +1,30 @@
...
-                workspace = workspace,
-                newDatasetName = "dataset-${UUID.randomUUID()}"
-            ) { infrastructure ->
+        val modification = AwsDatasetModification.Builder(
+            aws = aws,
+            dataset = sourceDataset
+        )
+            .workspace(workspace)
+            .build()
+        val transformation = object : DatasetTransformation {
```
dagguh: If we go back to `Consumer<Infrastructure<*>`, then we can:
```kotlin
modification.apply { infrastructure ->
    val jiraHome = infrastructure.jira.jiraHome
    val backupPath = "${jiraHome.location}/export"
    Ssh(jiraHome.host, connectivityPatience = 4)
        .newConnection()
        .use { ssh ->
            val listCommand = "ls -lh $backupPath"
            val listOutput = ssh.execute(listCommand).output
            println("$ $listCommand\n$listOutput")
            ssh.execute("rm -r $backupPath")
        }
}
```

wyrzyk: How it differs for `DatasetTransformation`?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
mcmurphy2000: Because of this discussion:
https://github.com/atlassian/aws-infrastructure/pull/30#discussion_r298488070

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
```diff
@@ -0,0 +1,136 @@
...
+        return storedDataset
+    }
+
+    private fun release(resource: Resource) {
+        logger.info("Releasing AWS resources ...")
+        resource.release().get(2, TimeUnit.MINUTES)
+        logger.info("AWS resources released")
+    }
+
+    private fun cleanUp(infrastructure: Infrastructure<*>) {
```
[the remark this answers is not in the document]
mcmurphy2000: I thought that we agreed that you will add this later. Maybe as a bugfix release

src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModificationIT.kt
wyrzyk: Do you mean it's more readable with `Consumer<Infrastructure<*>`?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
dagguh: Add the `transformation` as the `AwsDatasetModification` field. Then the `Builder` allows us to add `preStartTransformation` and `postStopTransformation` in the future.
We also get rid of the `Function<>`

dagguh: OTOH if we predict dependencies between these transformations, then we might be tempted to bind them together in the `DatasetTransformation`. TBH I'm fine with both, but if we're planning to add methods to interfaces, then we gotta port it to Java.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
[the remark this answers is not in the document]
dagguh: Alright, I'll add it in. In the meantime all users of this new API need to know that if they're modifying a public dataset, they can no longer publish the modified one.

src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModificationIT.kt
dagguh: It's more readable, because you don't need anonymous implementations, just lambdas.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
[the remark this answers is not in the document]
mcmurphy2000: Added a log entry to warn against publishing the modified dataset

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
mcmurphy2000: > Add the transformation as the AwsDatasetModification field. Then the Builder allows us to add preStartTransformation and postStopTransformation in the future.
We also get rid of the Function

That's exactly what I asked here:
https://github.com/atlassian/aws-infrastructure/pull/30#discussion_r298473750
And the conclusion was that "No-op seems not to be a good default"

wyrzyk: You can create a builder without default.

dagguh: The vessel for defaults (interface or builder) doesn't matter if there's no good default.
And if you want a new parameter/method, you have to either break compatibility or bring a default.

dagguh:
CI flaked due to JPERF-496, so we gotta bump infra to 4.12.4 to get the fix
But infra 4.12.4 shipped a new bug: JPERF-506, so we really gotta *bump infra to 4.12.6*

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
wyrzyk: I'm not sure if we are on the same page.
1. I'm ok with an interface. There are some problems with the interfaces/default methods and Java interop, but it's not a big deal. 
2. I'm ok with a builder. It will probably be more flexible, but we miss the one place we can inject a transformation that requires to be injected in multiple places of the infra's lifecycle.
3. I'm more or less ok with NoOp. It's not very bad, but the transformation with the default will be in most cases just a mistake. The builder can require a transformation, so it could eliminate a need for the default.

dagguh: ad. 1 Agreed
ad. 2 Agreed
ad. 3 Agreed. And if someone wants only to purge the dataset (e.g. me), they can easily pass in a NoOp themselves. We can provide a NoOp impl for them out-of-the-box to avoid copying the same boilerplate across clients.

PR #31, opened 2019-07-01
JPERF-509: Refactor `AwsDatasetModification`

mcmurphy2000:


src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
```diff
@@ -120,6 +120,7 @@ class AwsDatasetModification private constructor(
             aws = aws
         )
 
+        fun onlineTransformation(onlineTransformation: Consumer<Infrastructure<*>>) = apply { this.onlineTransformation = onlineTransformation }
```
dagguh: I like it
We should include the rationale in the commit message

dagguh: image
Few things here:
* the summary is too long and is being wrapped, imho "Refactor `AwsDatasetModification` summarizes the commit well enough
* `enchance` -> `enhance`
* `extention` -> `extension`

PR #36, opened 2019-07-05
Publish a small dataset

dagguh:


dagguh:
Hmm, this modifies the `AwsDatasetModification.kt`, but the `CODEOWNERS` didn't pick up @mcmurphy2000 
Perhaps I used the wrong notation

PR #47, opened 2019-08-23
JPERF-553: Provide a `NetworkFormula`

dagguh:


src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/Network.kt
```diff
@@ -1,8 +1,11 @@
-package com.atlassian.performance.tools.awsinfrastructure.api
+package com.atlassian.performance.tools.awsinfrastructure.api.network
```
dagguh: We can make the move, because `com.atlassian.performance.tools.awsinfrastructure.api.Network` class is not released yet

dagguh:
IDE failed to move `NetworkFormula`, force-pushed a compilation fix.
The CI saved my ass

PR #57, opened 2019-09-20
JPERF-579: Enables infrastructure formula to be created with user pro…

drspartez:
…visioned Jira

There were several situations when JPT tests should run over existing jira. It is possible through various hacks, which always end on not merged branches, then knowledge is lost.

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
@@ -3,6 +3,7 @@ package com.atlassian.performance.tools.awsinfrastructure.api
 import com.atlassian.performance.tools.aws.api.*
 import com.atlassian.performance.tools.awsinfrastructure.api.jira.DataCenterFormula
 import com.atlassian.performance.tools.awsinfrastructure.api.jira.JiraFormula
+import com.atlassian.performance.tools.awsinfrastructure.api.jira.ProvisionedJira
 import com.atlassian.performance.tools.awsinfrastructure.api.jira.StandaloneFormula
 import com.atlassian.performance.tools.awsinfrastructure.api.network.Network
 import com.atlassian.performance.tools.awsinfrastructure.api.network.NetworkFormula
@@ -27,12 +28,24 @@ import java.util.concurrent.Executors
  * - [Ec2VirtualUsersFormula]
  * - [MulticastVirtualUsersFormula]
  */
-class InfrastructureFormula<out T : VirtualUsers>(
+class InfrastructureFormula<out T : VirtualUsers> private constructor(
     private val investment: Investment,
-    private val jiraFormula: JiraFormula,
+    private val jiraFormula: JiraFormula?,
+    private val userProvisionedJira: ProvisionedJira?,
     private val virtualUsersFormula: VirtualUsersFormula<T>,
     private val aws: Aws
 ) {
+
+    constructor(investment: Investment,
+                userProvisionedJira: ProvisionedJira,
+                virtualUsersFormula: VirtualUsersFormula<T>,
+                aws: Aws) : this(investment, null, userProvisionedJira, virtualUsersFormula, aws)
+
+    constructor(investment: Investment,
+                jiraFormula: JiraFormula,
+                virtualUsersFormula: VirtualUsersFormula<T>,
+                aws: Aws) : this(investment, jiraFormula, null, virtualUsersFormula, aws)
+
     fun provision(
         workingDirectory: Path
     ): ProvisionedInfrastructure<T> {
@@ -59,8 +72,9 @@ class InfrastructureFormula<out T : VirtualUsers>(
         }
 
         val network = networkProvisioning.get()
-        val provisionJira = executor.submitWithLogContext("jira") {
-            overrideJiraNetwork(network).provision(
+
+        val provisionedJira =  userProvisionedJira ?: executor.submitWithLogContext("jira") {
+            overrideJiraNetwork(network)!!.provision(
                 investment = investment,
                 pluginsTransport = aws.jiraStorage(nonce),
                 resultsTransport = resultsStorage,
@@ -68,7 +82,7 @@ class InfrastructureFormula<out T : VirtualUsers>(
                 roleProfile = roleProfile,
                 aws = aws
             )
-        }
+        }.get()
 
         val provisionVirtualUsers = executor.submitWithLogContext("virtual users") {
             overrideVuNetwork(network).provision(
@@ -83,7 +97,6 @@ class InfrastructureFormula<out T : VirtualUsers>(
             )
         }
 
-        val provisionedJira = provisionJira.get()
         val provisionedVirtualUsers = provisionVirtualUsers.get()
         val sshKey = keyProvisioning.get()
 
@@ -108,7 +121,7 @@ class InfrastructureFormula<out T : VirtualUsers>(
 
     private fun overrideJiraNetwork(
         network: Network
-    ): JiraFormula = when (jiraFormula) {
+    ): JiraFormula? = when (jiraFormula) {
         is DataCenterFormula -> DataCenterFormula.Builder(jiraFormula).network(network).build()
         is StandaloneFormula -> StandaloneFormula.Builder(jiraFormula).network(network).build()
         else -> jiraFormula
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-2.17.0...master
 
+### Added
+- Expose API to create InfrastructureFormula with user provisioned Jira. Resolve [JPERF-579].
+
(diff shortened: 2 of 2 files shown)


dagguh:
We can simplify this a lot by implementing a `JiraFormula`, which accepts a `ProvisionedJira`. It'd be less intrusive WRT the existing API.

dagguh:
Furthermore we can make it even easier for this use case, by also implementing a `JiraFormula`, which accepts only an `URI`.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
```diff
@@ -2,7 +2,9 @@ package com.atlassian.performance.tools.awsinfrastructure.api
 
 import com.atlassian.performance.tools.aws.api.*
 import com.atlassian.performance.tools.awsinfrastructure.api.jira.DataCenterFormula
+import com.atlassian.performance.tools.awsinfrastructure.api.jira.Jira
```
mcmurphy2000: It seems this import is not needed

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
```diff
@@ -108,7 +122,7 @@ class InfrastructureFormula<out T : VirtualUsers>(
 
     private fun overrideJiraNetwork(
         network: Network
-    ): JiraFormula = when (jiraFormula) {
+    ): JiraFormula? = when (jiraFormula) {
```
lipinskipawel: Why you added `?`. Is it possible to avoid that change?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
```diff
@@ -59,16 +73,17 @@ class InfrastructureFormula<out T : VirtualUsers>(
         }
 
         val network = networkProvisioning.get()
-        val provisionJira = executor.submitWithLogContext("jira") {
-            overrideJiraNetwork(network).provision(
+
+        val provisionedJira = if (userProvisionedJira != null) userProvisionedJira else executor.submitWithLogContext("jira") {
```
mcmurphy2000: this can be simplified as
`val provisionedJira = userProvisionedJira ?: executor.submitWithLogContext("jira")`

drspartez:
> We can simplify this a lot by implementing a `JiraFormula`, which accepts a `ProvisionedJira`. It'd be less intrusive WRT the existing API.
> Furthermore we can make it even easier for this use case, by also implementing a JiraFormula, which accepts only an URI.
With this change we don't need to implement anything more. Just use Jira class with valid URI and resource addresses if needed. So it is simple and the change doesn't break API.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
drspartez: Yeahhh- Elvis :)

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
drspartez: Because jira formula might be null. I wan't it to be avoided because there are two options to deal with jira (by formula or by already provisioned stuff)

dagguh:
We can provide that implementation as API. So the caller still doesn't have to implement anything. But the diff here would be completely unintrusive.

drspartez:
> We can provide that implementation as API. So the caller still doesn't have to implement anything. But the diff here would be completely unintrusive.

I believe we could achieve that with 100+ ways. This is simple (KISS) and gives the api value which is in demand. 
It does not introduce any new artifacts which is good because serves simplicity.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
```diff
@@ -59,16 +72,17 @@ class InfrastructureFormula<out T : VirtualUsers>(
         }
 
         val network = networkProvisioning.get()
-        val provisionJira = executor.submitWithLogContext("jira") {
-            overrideJiraNetwork(network).provision(
+
+        val provisionedJira =  userProvisionedJira ?: executor.submitWithLogContext("jira") {
```
dagguh: Please explain how is this simpler than https://github.com/atlassian/aws-infrastructure/pull/58

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
```diff
@@ -59,16 +72,17 @@ class InfrastructureFormula<out T : VirtualUsers>(
...
+            overrideJiraNetwork(network)!!.provision(
```
dagguh: Now it started throwing NPEs

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
drspartez: It is one more block in API. Feel free to continue this way. I just need to have one way to  achieve that.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
drspartez: public constructors prevents NPE here

dagguh: Indeed the constructors plus the elvis null-check might indeed prevent it. It's a few moving parts working together to prevent the NPE.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
dagguh: I get it, it's choosing the `JiraFormula` impl vs choosing the `InfrastructureFormula` constructor. Neither is strictly simpler than the other.
*But* the API already forces the choice of `JiraFormula` impl, so it fits the current patterns. And also this impl requires internal null checking and additional branches of execution (hidden `if`s) and more internal states to track and debug.
Also the parameter-expanding design is a poor trajectory for API evolution, because there's more and more permutations of optional parameters. Imagine adding another optional feature, now we'd have 4 constructors, then 8.

dagguh: So I hope you're aware of the intricacies, alternatives and hidden costs. I'm gonna unblock the PR

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
```diff
@@ -27,12 +28,24 @@ import java.util.concurrent.Executors
  * - [Ec2VirtualUsersFormula]
  * - [MulticastVirtualUsersFormula]
  */
-class InfrastructureFormula<out T : VirtualUsers>(
+class InfrastructureFormula<out T : VirtualUsers> private constructor(
```
dagguh: Thanks for keeping the internals hidden

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
```diff
@@ -59,16 +72,17 @@ class InfrastructureFormula<out T : VirtualUsers>(
...
+            overrideJiraNetwork(network)!!.provision(
                 investment = investment,
                 pluginsTransport = aws.jiraStorage(nonce),
                 resultsTransport = resultsStorage,
                 key = keyProvisioning,
                 roleProfile = roleProfile,
                 aws = aws
             )
-        }
+        }.get()
```
dagguh: This change breaks the concurrency of Jira and VU provisioning. We don't have a test for `InfrastructureFormula` so CI doesn't catch this at the moment. We gotta revert this

dagguh: PS. such test is included in a [WIP side-branch](https://github.com/atlassian/aws-infrastructure/compare/master...dagguh:issue/JPERF-554-kibana-results#diff-19cab7e2eed0cf07a440956071c59558)

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
dagguh: Given the choice, you merged this version, despite all of the pros and cons. Could you share the rationale?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
drspartez: I'm going to fix that.

PR #58, opened 2019-09-20
JPERF-579: Add `ProvisionedJiraFormula`

dagguh:
Alternative to https://github.com/atlassian/aws-infrastructure/pull/57
No nulls, no NPE checks, no elvises, no extra constructors.
Still keeps feature set open to further extension without generating permutations of optional parameters.

PR #59, opened 2019-09-20
Issue/jperf 579 just give me an uri

dagguh:
This goes one step further than https://github.com/atlassian/aws-infrastructure/pull/58 - simply pass in an `URI`.


dagguh:
This is a version, where the user really doesn't have to do much - just an `URI`.

dagguh:
Both options are trivial to implement on the caller side. They can be used and tested without merging this PR. When it is proven to work well, we can contribute it here.

dagguh:
Pushed in a rebase, and a revert. See the rationale in 43e884ba3c29002c223ecd96c62d5049c22e3121

PR #60, opened 2019-09-24
JPERF-579: Enables infrastructure formula to be created with user pro…

drspartez:
…visioned Jira

There were several situations when JPT tests should run over existing jira. It is possible through various hacks, which always end on not merged branches, then knowledge is lost.

It also reverts previous commit related to this issue and brings proper concurrency to VU and Jira provisioning jobs.


src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
```diff
@@ -97,6 +96,7 @@ class InfrastructureFormula<out T : VirtualUsers> private constructor(
             )
         }
 
+        val provisionedJira = provisionJira.get()
```
sebapawlak: Not sure how the integration test linked by Maciej in the previous PR works but perhaps we would want to copy-paste it to this PR and have something that guards this change. WDYT?

drspartez: Seems to be good hint - i need to investigate what is inside this IT

PR #63, opened 2019-11-21
parameterise the instance type for virtual users

pbruski:


src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/virtualusers/MulticastVirtualUsersFormula.kt
```diff
@@ -20,7 +21,8 @@ class MulticastVirtualUsersFormula private constructor(
     private val nodes: Int,
     private val splunkForwarder: SplunkForwarder,
     private val browser: Browser,
-    private val network: Network?
+    private val network: Network?,
+    private val instanceType: InstanceType
```
dagguh: It's not used.
You prolly want to pass it into the `StackVirtualUsersFormula.Builder`.

pbruski: Damn. Fixed!

PR #65, opened 2019-11-22
Mark 2.19.0

dagguh:


dagguh:
@pbruski it's available in 2.19.0 now

PR #66, opened 2019-11-26
Fix copying builders to include the instance type parameter too

pbruski:


src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/virtualusers/Ec2VirtualUsersFormula.kt
```diff
@@ -110,6 +110,7 @@ class Ec2VirtualUsersFormula private constructor(
             browser = formula.browser
             network = formula.network
             nodeOrder = formula.nodeOrder
+            instanceType = formula.instanceType
```
dagguh: Ah, we've been hit with copying builders before: covered in 56f293cd5abdbbe7d762a1a32b864c3e68900e1e and fixed in 49be8499728d6b6e8bddcfcb6f0903fb1441437a, but not asserted.
Assertions of these internals are tricky in integration tests. It should be feasible in unit tests. But I won't require a TDD red-green-refactor commit structure here.

pbruski:
> We're going for a linear (or at least acyclic) Git history, so we rebase onto `upstream/mater` rather than merging. I can cherry-pick the PR, and rebase it and create a new one, declining this one. Or you can rebase it yourself.

I've opened a new PR

PR #67, opened 2019-11-27
fix copying builders to include the instance type parameter too

pbruski:


dagguh:
Thanks for the clean Git graph!

PR #68, opened 2019-11-27
Support private ip addresses

pbruski:


dagguh:
If you want to clean up the Git graph, you can force-push to your branch. Thanks to forks, you have to full admin, sudo, God-mode powers over your repo, so you can force-push freely 😉

PR #71, opened 2019-11-27
allow the use of pre-provisioned networks for VUs

pbruski:


src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
```diff
@@ -31,8 +31,16 @@ class InfrastructureFormula<out T : VirtualUsers>(
     private val investment: Investment,
     private val jiraFormula: JiraFormula,
     private val virtualUsersFormula: VirtualUsersFormula<T>,
-    private val aws: Aws
+    private val aws: Aws,
+    private var preProvisionedNetwork: Network?
```
dagguh: Let's use `val` (Java `final`) by default. Mutable references should be the exception.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
```diff
@@ -31,8 +31,16 @@ class InfrastructureFormula<out T : VirtualUsers>(
...
 ) {
+    constructor(
```
dagguh: Multiple constructors mean it's time to go for a builder instead. See [Java API checklist 3.6.12](https://github.com/wyrzyk/java-api-design-checklist/blob/d0dbffa525007f320892b382a3e477702707cd93/CHECKLIST.md#3612-consider-a-builder-when-faced-with-many-constructor-parameters). Then, the new constructor should be `private` to shield maintainers from future compatibility pain.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
dagguh: The nullability here represents optional behavior (provisioning a network or not).
Perhaps we could replace it with bona fide polymorphism. E.g. `Supplier<Network>`. We'd need such [indirection anyway, if we go for builders](https://github.com/wyrzyk/java-api-design-checklist/pull/10).

dagguh: Hmm, we won't need it for builders if the default is `null`, fair enough

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
pbruski: added

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
```diff
@@ -123,6 +143,25 @@ class InfrastructureFormula<out T : VirtualUsers>(
         is MulticastVirtualUsersFormula -> MulticastVirtualUsersFormula.Builder(virtualUsersFormula).network(network).build()
         else -> virtualUsersFormula
     } as VirtualUsersFormula<T>
+
+    class Builder<out T : VirtualUsers>(
+        private val investment: Investment,
+        private val jiraFormula: JiraFormula,
+        private val virtualUsersFormula: VirtualUsersFormula<T>,
+        private val aws: Aws
```
dagguh: The more required parameters, the more it hurts to use it 🤕.
We can add sensible defaults for everything except for `aws` (due to its impure constructor).
E.g. we can use `StandaloneFormula`, `StackVirtualUsersFormula` and a generic `Investment`.
Providing a `StandaloneFormula` hurts a bit due to the 3 params (🤕). It could easily have a zero-arg builder.
Providing a SVUF hurts only a little, due to 1 mandatory param. But there's no sensible default, so we'll have to ask for this param too.

All of the above can be done later, by providing a new `InfrastructureFormula.Builder` constructor, so I'm ok with merging as is, so it's your call.

pbruski: It's problematic with SVUF because it's generic. But I've added the others.

PR #73, opened 2019-11-28
a workaround for self-provisioned Jira nodes

pbruski:


src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/Jira.kt
```diff
@@ -16,9 +18,15 @@ class Jira(
     val address: URI,
     val jmxClients: List<JmxClient> = emptyList()
 ) : MeasurementSource {
+    private val logger: Logger = LogManager.getLogger(this::class.java)
+
     override fun gatherResults() {
+        if (nodes.isEmpty()) {
```
dagguh: Yikes, good catch. `Executors.newFixedThreadPool` `@throws IllegalArgumentException if {@code nThreads <= 0}`
That's my bad for skipping tests in https://github.com/atlassian/aws-infrastructure/pull/59

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/Jira.kt
```diff
@@ -16,9 +18,15 @@ class Jira(
...
+            logger.warn("No Jira nodes known to JPT, not downloading node results")
+            return
+        }
         val executor = Executors.newFixedThreadPool(
-            Math.min(nodes.size, 4),
+            nodes.size.coerceAtMost(4),
```
dagguh: Ah yeah, our ancient non-idiomatic premature microoptimization :)
Alternatively, we could simplify away the thread pool size control (and the `if`) via `ExecutorService.newCachedThreadPool`

PR #76, opened 2019-12-11
avoid sudo warning about unresolved hosts

pbruski:


dagguh:
Thanks for the fix!
This is an ancient bug we were tracking since May 2018. I raised a [public ticket](https://ecosystem.atlassian.net/browse/JPERF-595) right now.
Could you add an entry in the `CHANGELOG.md`?

PR #79, opened 2020-01-31
vu ami update

pbruski:


src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/virtualusers/AmiNameResolver.kt
```diff
@@ -0,0 +1,49 @@
+package com.atlassian.performance.tools.awsinfrastructure.api.virtualusers
+
+import com.amazonaws.services.ec2.model.DescribeImagesRequest
+import com.amazonaws.services.ec2.model.Filter
+import com.atlassian.performance.tools.aws.api.Aws
+import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.ConcurrentMap
+
+internal class AmiNameResolver {
```
dagguh: `internal` shouldn't be in `api`

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/virtualusers/StackVirtualUsersFormula.kt
```diff
@@ -89,9 +89,11 @@ class StackVirtualUsersFormula private constructor(
     ): ProvisionedVirtualUsers<SshVirtualUsers> {
         logger.debug("Setting up $name...")
         val network = overriddenNetwork ?: NetworkFormula(investment, aws).provision()
+        val cloudformationTemplate = readResourceText("aws/virtual-users.yaml")
+        logger.info("template: $cloudformationTemplate")
```
dagguh: What is it for? Isn't it the same as reading from source code directly?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/virtualusers/AmiNameResolver.kt
pbruski: moved out

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/virtualusers/StackVirtualUsersFormula.kt
pbruski: debug, removing

PR #80, opened 2020-02-04
HL-1205 Postgres support

masiorski:


CHANGELOG.md
```diff
@@ -21,7 +21,16 @@ Adding a requirement of a major version of a dependency is breaking a contract.
 Dropping a requirement of a major version of a dependency is a new contract.
 
 ## [Unreleased]
-[Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-2.20.0...master
+[Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-2.20.2...master
+
+## [2.20.2] - 2020-02-04
```
wyrzyk: > Given a version number MAJOR.MINOR.PATCH, increment the:
> 
> MAJOR version when you make incompatible API changes,
> **MINOR version when you add functionality in a backwards compatible** manner, and

https://semver.org/

Shouldn't it be 2.21.0?

masiorski: it can be sir :) I will change this

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneNodeFormula.kt
```diff
@@ -114,6 +115,39 @@ internal class StandaloneNodeFormula(
         connection: SshConnection,
         dbconfigXml: String
     ) {
+        if (!isMySql) {
+            connection.execute("sudo rm -f $dbconfigXml")
+
+            var dbconfigContent = """
+<?xml version="1.0" encoding="UTF-8"?>
```
dagguh: We can add the XML to `src/main/resources` and read it from classpath.
This way we'll be able to see XML with proper syntax highlight and verification.

src/main/resources/aws/2-nodes-dc.yaml
```diff
@@ -104,6 +104,7 @@ Resources:
       SecurityGroupIds:
         - Ref: SshSecurityGroup
         - Ref: MySqlSecurityGroup
+        - Ref: PostgresSecurityGroup
```
dagguh: It's a bit shoehorned, but shouldn't have negative consequences AFAIK

src/main/resources/aws/2-nodes-dc.yaml
```diff
@@ -183,6 +184,17 @@ Resources:
           FromPort: 3306
           ToPort: 3306
           CidrIp: 0.0.0.0/0
+  PostgresSecurityGroup:
+    Type: AWS::EC2::SecurityGroup
```
dagguh: The mini-disadvantage is that it will allocate two Security Groups even tho only one is used. It's an AWS resource, so it'll cost a little and might increase CloudFormation provisioning time. OTOH SecurityGroups are relatively cheap, so I guess we're fine here. Just FYI.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneNodeFormula.kt
```diff
@@ -114,6 +115,39 @@ internal class StandaloneNodeFormula(
         connection: SshConnection,
         dbconfigXml: String
     ) {
+        if (!isMySql) {
+            connection.execute("sudo rm -f $dbconfigXml")
```
dagguh: It nicely explains the current tribal knowledge: the MySQL datasets have predefined `dbconfig.xml` 👍

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/StandaloneFormula.kt
```diff
@@ -180,6 +181,7 @@ class StandaloneFormula private constructor(
         CloseableThreadContext.push("Jira node").use {
             key.get().file.facilitateSsh(jiraIp)
         }
+        val isMySql = database is MySqlDatabase
```
dagguh: This will cause `Database` decorators to malfunction. [They do exist in the wild](https://github.com/atlassian/jira-hardware-exploration/blob/a8395bec1f96a36a4a5f4e03527e071532c932ae/src/test/kotlin/com/atlassian/performance/tools/lib/infrastructure/ConfigurableMysqlDatabase.kt).
In that case, `isMySql` will be `false` and subsequently we'll delete `dbconfig.xml` and break that use case.

So let's detect `is PostgresDatabase` instead, because such callers don't exist yet. This is a short-term hack, so we can still 🤞 for the [infra/aws-infra fix](https://github.com/atlassian/infrastructure/pull/43).

- [ ] detect `is PostgresDatabase` instead

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneNodeFormula.kt
masiorski: ok, I'll do this

PR #82, opened 2020-02-19
JPERF-188: Improve stopJira to wait for Jira process to shutdown inst…

KonradMarszalek:
JPERF-188: Improve stopJira to wait for Jira process to shutdown instead of all Java processes on the node

Renamed branch and commit

Added Changelog entry

PR #84, opened 2020-02-20
JPERF-188: Improve stopJira to wait for Jira process to shutdown inst…

mzyromski-atlassian:
…ead of all Java processes on the node

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/CustomDatasetSource.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/CustomDatasetSource.kt
@@ -88,8 +88,10 @@ class CustomDatasetSource(
 
     private fun stopJira(host: SshHost) {
         val shutdownTomcat = "echo SHUTDOWN | nc localhost 8005"
-        val waitForNoJavaProcess = "while ps -C java -o pid=; do sleep 5; done"
-        Ssh(host, connectivityPatience = 4).newConnection().use { it.safeExecute("$shutdownTomcat && $waitForNoJavaProcess", Duration.ofMinutes(3)) }
+        val waitForNoJiraProcess = "while pgrep -f jira; do sleep 5; done"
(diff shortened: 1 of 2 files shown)


dagguh:
This is https://github.com/atlassian/aws-infrastructure/pull/82 adopted by @mzyromski-atlassian to cover the CI.
[CI shows](https://app.circleci.com/jobs/github/mzyromski-atlassian/aws-infrastructure/7/tests), that the change introduces an infinite Bash loop, which times out:
```
java.lang.Exception: SSH command failed to finish in extended time (PT3M45S): SshExecutedCommand(cmd=echo SHUTDOWN | nc localhost 8005 && while pgrep -f jira; do sleep 5; done, stdout=6481
7459
6481
7459
6481
7459
6481
7459
...
```

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/CustomDatasetSource.kt
```diff
@@ -88,8 +88,10 @@ class CustomDatasetSource(
 
     private fun stopJira(host: SshHost) {
         val shutdownTomcat = "echo SHUTDOWN | nc localhost 8005"
-        val waitForNoJavaProcess = "while ps -C java -o pid=; do sleep 5; done"
-        Ssh(host, connectivityPatience = 4).newConnection().use { it.safeExecute("$shutdownTomcat && $waitForNoJavaProcess", Duration.ofMinutes(3)) }
+        val waitForNoJiraProcess = "while pgrep -f jira; do sleep 5; done"
```
dagguh: Gotta fix that infinite loop from CI

KonradMarszalek:
It seems there are more processes with `jira` fragment in it's execution path in the CI. Let me improve the logging first so we can see what the processes are. Then we can narrow the pgrep query even more.

dagguh:
Alternatively we could read PID from the PID file.

PR #86, opened 2020-02-24
JPERF-235 set instance initiated shutdown behavior to terminate

pbruski:


dagguh:
[CI flake is known](https://ecosystem.atlassian.net/browse/JPERF-596?jql=project%20%3D%20JPERF%20AND%20text%20~%20%22%5C%22No%20such%20process%5C%22%22).

src/main/resources/aws/2-nodes-dc.yaml
```diff
@@ -36,6 +36,7 @@ Resources:
       KeyName:
         Ref: KeyName
       ImageId: !Ref Ami
+      InstanceInitiatedShutdownBehavior: terminate
```
dagguh: This behavior is significant. We want to rely on it externally. Therefore we want it to be a behavioral contract (KeepAChangelog `Added`, SemVer MINOR). [We specify behavioral contracts via Javadoc](https://github.com/atlassian/aws-infrastructure/commit/568f949ca1f4f4ec5f6592adab42592ba094061f#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR12). So let's add this behavior to:
* `DataCenterFormula`
* `StandaloneFormula`
* `StackVirtualUsersFormula`

While we think of the public API, let's not forget about its other parts, not covered yet:
* `Ec2VirtualUsersFormula`
* `ApacheEc2LoadBalancerFormula`

Also, tests are still "vulnerable": `AwsCliIT`
The common point for the non-Cloudformation ones is `aws.awaitingEc2.allocateInstance`, perhaps we could put the terminate as default there.

pbruski: aws.awaitingEc2.allocateInstance - maybe we should make the change in both places?

pbruski: i.e. here and in aws-resource

PR #90, opened 2020-05-12
JPERF-637: Do not specify availability zones when requesting AWS stack creation

mzyromski-atlassian:
Rely on internal allocation instead.

This is, as per AWS support folks, a preferred way to decrease frequency of Insufficient Capacity Errors that we've been observing quite a lot recently.


src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/IntegrationTestRuntime.kt
```diff
@@ -14,7 +14,6 @@ object IntegrationTestRuntime {
     init {
         ConfigurationFactory.setConfigurationFactory(LogConfigurationFactory(taskWorkspace))
         aws = Aws.Builder(Regions.EU_WEST_1)
-            .availabilityZoneFilter(Predicate { it.zoneName !in listOf("eu-west-1a") })
```
wyrzyk, line 17: I don't remember why we didn't change it before. AFAIR we wanted to provision all the instances in the same AZ to ensure both experiment and baseline run on similar network topology. CC: @dagguh

mzyromski-atlassian: Are you referring here to this particular changed line or to the entire changeset?

Currently we're not running baseline and experiment on the same AZ (it's randomly chosen) and my commit doesn't change it -- we'll still have different zones, but they will be assigned by AWS infra and not specified by us. Ensuring that we run everything on the same AZ is a different story and would require going even further in the initial direction which is considered by AWS folks a bad idea.

I propose to go with my change and see if it helps in terms of stability. If not, then we'll have an argument when opening a support case.

dagguh: The reason is in this diff in this PR:
https://github.com/atlassian/aws-infrastructure/pull/90/files#diff-a54501e4223a91aed36123deeaa1077dL8-L9

> Randomizing helps spread resources over a wider capacity pool.

If it actually doesn't help with capacity and provisioning, then we should indeed remove it.

Consistent AZs are not enforced anyway. We only try to keep the same `Network` across the `InfrastructureFormula`.

src/main/resources/aws/network.yaml
```diff
@@ -16,7 +13,6 @@ Resources:
       MapPublicIpOnLaunch: true
       CidrBlock: 10.0.0.0/24
       VpcId: !Ref Vpc
-      AvailabilityZone: !Ref AvailabilityZone
```
dagguh, line 19: I'd add `Fixed` changelog

mzyromski-atlassian: True! Thanks for reminding.

CHANGELOG.md
```diff
@@ -23,6 +23,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-2.21.0...master
 
+### Fixed
+- Do not explicitly specify Availability Zone when requesting resources to reduce insufficient capacity errors. Resolve [JPERF-637].
```
dagguh, line 27: Nit, a purely stylistic remark:
I try to specify what the code *does* rather than what it *doesn't do*.
E.g.:
```suggestion
- Let AWS pick the Availability Zone to balance resource capacity. Resolve [JPERF-637].
```

mzyromski-atlassian:
Released in `2.21.1`.

PR #92, opened 2020-05-19
JPERF-644: Update CLA links

mzyromski-atlassian:


mzyromski-atlassian:
Force-pushed same commit to trigger CI.

PR #96, opened 2020-06-23
do not log the worrying error

pbruski:


mzyromski-atlassian:
Released in `2.21.3`.

PR #100, opened 2020-07-10
Upgraded AWS SDK to 1.11.817

pbruski:


gradle/dependency-locks/apiDependenciesMetadata.lockfile
```diff
@@ -1,18 +1,18 @@
...
+com.amazonaws:aws-java-sdk-ec2:1.11.817
+com.amazonaws:aws-java-sdk-elasticloadbalancing:1.11.817
+com.amazonaws:aws-java-sdk-iam:1.11.817
+com.amazonaws:aws-java-sdk-kms:1.11.817
+com.amazonaws:aws-java-sdk-rds:1.11.817
+com.amazonaws:aws-java-sdk-s3:1.11.817
+com.amazonaws:aws-java-sdk-sts:1.11.817
+com.amazonaws:aws-java-sdk-support:1.11.817
+com.amazonaws:jmespath-java:1.11.817
+com.atlassian.performance.tools:aws-resources:1.6.1-SNAPSHOT
```
wyrzyk, line 15: Snapshot?

build.gradle.kts
```diff
@@ -27,6 +27,7 @@ configurations.all {
                 "com.fasterxml.jackson.core:jackson-core" -> useVersion("2.9.4")
                 "org.slf4j:slf4j-api" -> useVersion("1.8.0-alpha2")
                 "org.apache.httpcomponents:httpcore" -> useVersion("4.4.9")
+                "commons-codec:commons-codec" -> useVersion("1.11")
```
wyrzyk, line 30: A new dependency? What is it for?

wyrzyk: Ok, it's an old one. Just a version bump.

dagguh: Old dependency, new version conflict. That's what happens if libs don't use version ranges: uncountable bogus version conflicts.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/AmiNameResolver.kt
```diff
@@ -26,24 +43,37 @@ internal class AmiNameResolver {
...
 
-    data class AwsAndAmi constructor(val aws: Aws, val ami: String) {
+        private fun getVuAmiName(architecture: String): String {
+            val imgArchitecture = if (architecture != "arm64") {
+                "amd64"
+            } else {
+                architecture
+            }
+
+            return "ubuntu-focal-20.04-${imgArchitecture}-server-20200701"
```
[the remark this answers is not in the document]
pbruski: Hmmm. I've just realised that uses of aws-infrastructure will be broken if they use SDK older than the one where they introduced DescribeInstanceTypes. How do we handle that? That means a new major release, right?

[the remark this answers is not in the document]
dagguh: That's the problem with hiding version conflicts by default (damn you Maven). We can express that aws-infra requires at least this version, but it could be ignored.
Gradle consumers are in the clear. They can `failOnVersionConflict` to detect it early. And they can use dependency locks to control when the bump (and conflict) happens.
I tried to depend on AWS SDK SemVer range, but Maven would download every single POM for every version in the range (hundreds). It increased Maven build times by tens of minutes on cold caches. Gradle did no such thing.
Our examples for vendors use Maven, because the Atlassian ecosystem is built upon it. 

So in general, it's not a problem for Gradlers, and it might be a problem for Maveners.
Even the loose Gradle users (without dep locks and strict conflicts) would not be impacted, because Gradle would pick the higher version anyway. Maven will pick the closest dep in the tree.

By accident, `jira-performance-tests` depends on `aws-infrastructure` first and then on `aws-resources`, so aws-infra AWS SDK will win. If we ever need to take the control of this, `jira-performance-test` can resolve this conflict in its POM.

So for now, we can stick to the [current POM contracts](https://github.com/atlassian/aws-infrastructure/blob/37f70fc32fa801cbe7df73a2921b5d79e80544c0/CHANGELOG.md#pom):
> Adding a requirement of a major version of a dependency is breaking a contract. Dropping a requirement of a major version of a dependency is a new contract.

If we ever find a way to (any of the following):
* get `failOnVersionConflict` for Maven
* get dependency locks for Maven (this is actually very probable)
* get Maven to stop downloading the entire dependency version range
* avoid Maven as a build tool

... then we can rest easy with the current POM contracts.
A better POM contract could help us too, but I'm out of ideas there.

PR #107, opened 2020-12-03
JPERF-632: Make mkfs.ext4 execution quiet

mzyromski-atlassian:
This will get rid of the command printing out version number string (e.g. "mke2fs 1.42.13 (17-May-2015)") to stderr which then gets populated to logs as a warning. -q does not prevent the command from printing out to stderr actual errors though.

Also, remove unused, duplicated ephemeral storage creation code.


dagguh (commented):
Nice, it's been making noise for too long 🙂

dagguh (approved):

PR #109, opened 2021-04-07
JPERF-3318 Enable EBS volume encryption

dmewada-atlas:
The jces-1209 (JPT for cloud) will be running for EAP (Early access program) customer and as per the security team's suggestions, we need to encrypt the EBS volume.
I think I have already shared our solution plan in #help-jpt channel, but let me know if you need it then will share it.
Thanks

### DIFF
--- a/src/main/resources/aws/virtual-users.yaml
+++ b/src/main/resources/aws/virtual-users.yaml
@@ -39,6 +39,7 @@ Resources:
           Ebs:
             VolumeSize: 30
             VolumeType: gp2
+            Encrypted: true
   SshSecurityGroup:
     Type: AWS::EC2::SecurityGroup
     Properties:


wyrzyk (approved):

src/main/resources/aws/virtual-users.yaml
```diff
@@ -39,6 +39,7 @@ Resources:
           Ebs:
             VolumeSize: 30
             VolumeType: gp2
+            Encrypted: true
```
dagguh, line 42: @mzyromski-atlassian FYI this was missing a changelog entry

mzyromski-atlassian: Thanks for the info. I'll include it in the changelog after this gets released.

mzyromski-atlassian:
Released in `2.23.0`.

PR #116, opened 2021-10-27
JPERF-730: Restrict access to AWS instances provisioned by JPT only to the necessery connections

pczuj:
If connection from the outside of VPC is needed (e.g. for debugging or jmx) it's recommended to use SSH tunnel.


mzyromski-atlassian (dismissed):

pczuj:
Unfortunately this is not final. It's very close to be finished, but we need to test something first.

pczuj:
This PR is now ready for review.

I'm aware that not everything is perfect in those changes:
 * Main commit contains multiple changes - could be split into "preparation" and "the change"
 * Unit tests for new classes are missing
 * There is no intergation test for the external access limitation
 * Some changes are unnecessary
 * Changes are not backwards compatible - this could be done in 2 steps, so breaking would be only about changing the defaults + eventual removal of deprecated methods

Please let me know if anything from the list or not mentioned is important to implement before proceeding with the merge.

pczuj:
Summary of each change included in this PR:
 * Bumped aws sdk and httpclient dependency versions to make it compatible with recent versions of `aws-resources`, which contains new API necessery e.g. for changes in `ApacheEc2LoadBalancerFormula` or in `Ec2VirtualUsersFormula`. 
 * Generated new dependency locks.
 * Added JVM memory settings in `gradle.properties` because on my machine I encountered some errors without it.
 * Changed all cloudformation templates so that they are limiting external access to provisioned resources completely.
 * Wherever it was possible made the cloud formation stacks to communicate though internal AWS network rather than public.
 * Explained why do we need full port access between Jira nodes (Jira 7.2 compatibility).
 * (Unnecessary change) Changed security groups to model groups of similar instances rather than port access thay provide. Updated descriptions to match this model.
 * (Unnecessary change) Extracted from `TemplateBuilder` Jira DC specific implementation (`JiraTemplateBuilder`). 
 * `TemplateBuilder` can still add security group ingress rules, however it won't do it automatically for jvm debug, jmx and splunk forwarder when using `JiraTemplateBuilder#adaptTo`. New mechanism for opening ports is introduced - will be mentioned in later bullet point. 
 * (Unnecessary change) Added DEBUG logging of generated template content after calling `build()` of `TemplateBuilder`
 * (Unnecessary change) Added better ERROR logging for failures happening during Jira startup by implementing `DiagnosableStoppedNode` and using it in `DiagnosableNodeFormula`.
 * (Unnecessary change) Changed `LoadBalancerFormula` interface to accept `Network` instead of `Vpc` and `Subnet` separately. `Network` = `Vpc` + `Subnet`.
 * (Unnecessary change) Suppressed warnings for builders and duplicated code between `DataCenterFormula` and `StandaloneFormula`.
 * (Unnecessary change) Deprecated `InfrastructureFormula` constructor. It uses a builder.
 * (Unnecessary change) Added `Builder` to `Jira` class and made every field but `url` optional. This broke backward compatibility, as `jiraHome` can now be `null`, however IMHO it makes sense this way. It will help with implementation of solution for JPERF-729.
 * Extracted `Provisioned...` classes (Jira, LoadBalancer, VirtualUsers, Infrastructure) to separate files.
 * Added `Builder`s to `Provisioned...` classes. 
 * Introduced new `Defaults` pattern to share default values between deprecated constructors, builder and lib consumers, who'd like to reuse the builder's default, e.g. to wrap a default value in additional functionality like `MultiAccessRequester`.
 * (Actual change) Added `AccessProvider` and `AccessRequester` abstraction. Provisioned components can communicate using it to share access. `AccessProvider` can define how the access is added (modifying stack or adding ingress rules), while `AccessRequester` by consuming the provider can decide to which IP subnet or security group the access should be granted. e.g. in `DataCenterFormula` for http Jira nodes expose `AccessProvider` which is consumed by load balancer's `AccessRequester`. Load balancer also exposes `AccessProvider` which is consumed by both virtual user's `AccessRequester` and external `AccessRequester`. Access to ports other than http is provided similarly.
 * Added `AccessProvider` and `AccessRequester` as fields to `Provisioned...` classes.
 * (Unnecessary change) Refactored fields for IP and SSH in formula classes to be more verbose about their real identity, e.g. private or public IP.
 * (Unnecessary change) Refactored `StandaloneFormula` a bit to match `DataCenterFormula` order of step execution.
 * (Unnecessary change) Made Jira setup in `StandaloneFormula` asynchronous.
 * Added `AccessRequester` as field for `DataCenterFormula` and `StandaloneFormula` building. This way there can be a default.
 * Created a default implementation of `AccessRequester` for `DataCenterFormula` and `StandaloneFormula`, which will pull the public IP address of the test coordinator machine from the internet (currently from an undocumented AWS service) and request to add that IP to whitelist for access to useful Jira ports (http and if enabled: jmx, jvm debug, splunk forwarding).
 * Extended INFO log line about Jira being started to mention that external access to the instance is restricted.

mzyromski-atlassian (commented):
How about adding, together with the changes, CHANGELOG entry that describes what's new. As the change is important and will require major build we could make the entry a bit longer than usual.

pczuj:
I'm splitting the changes into separate commits to indicate them in history. 

I'll drop 2 that are breaking: "`Jira.Builder`" and usage of `Network` in `LoadBalancerFormula`. Possibly will add them as optional commits at the end. If we are doing a major release it could be still good to merge those.

I'll also prepare a proposal for CHANGELOG entries.

pczuj:
Ops... I broke one test by removing a method and not transitioning to a new one. Correcting that...

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
```diff
@@ -281,14 +350,15 @@ class DataCenterFormula private constructor(
             address = loadBalancer.uri,
             jmxClients = jiraNodes.mapIndexed { i, node -> configs[i].remoteJmx.getClient(node.publicIpAddress) }
         )
-        logger.info("$jira is set up, will expire ${jiraStack.expiry}")
+        logger.info("$jira is set up, will expire ${jiraStack.expiry}. External access to the application is restricted with firewall (See JPERF-730).")
```
mzyromski-atlassian, line 353: Issue key might not say much to non-Atlassian users. How about having an URL to the tickets instead?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
```diff
@@ -0,0 +1,61 @@
...
+        .mapNotNull { serviceUrl -> urlReader(serviceUrl)?.trim() }
+        .filter { it.isNotEmpty() }
+        .filter { ipValidator(it) }
+        .toSet()
+        .let { set ->
+            if (set.isEmpty()) {
+                throw IllegalStateException("Queried services didn't report any valid public IP for this machine")
+            }
+            if (set.size != 1) {
+                throw IllegalStateException("Queried services reported different public IPs for this machine: ${set.toList()}")
```
mzyromski-atlassian, line 40: different -> multiple? Also, I'm not sure if this message is actionable for the user. What are they supposed to do in such a case?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
```diff
@@ -0,0 +1,61 @@
...
+
+    override fun get() = servicesToQuery
+        .asSequence()
+        .mapNotNull { serviceUrl -> urlReader(serviceUrl)?.trim() }
+        .filter { it.isNotEmpty() }
+        .filter { ipValidator(it) }
+        .toSet()
+        .let { set ->
+            if (set.isEmpty()) {
+                throw IllegalStateException("Queried services didn't report any valid public IP for this machine")
```
mzyromski-atlassian, line 37: I wonder, should we fail or just warn and go on? As I understand if we continue, the client machine won't get access non-SSH access to JPT nodes, which in many (most?) cases won't be an issue.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/UnableAccessProvider.kt
```diff
@@ -0,0 +1,13 @@
+package com.atlassian.performance.tools.awsinfrastructure.api.network.access
+
+import com.amazonaws.services.ec2.model.SecurityGroup
+
+class UnableAccessProvider : AccessProvider {
```
mzyromski-atlassian, line 5: How about renaming it to NoAccessProvider? Unless I'm getting the intent wrong here.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
```diff
@@ -238,6 +246,57 @@ class DataCenterFormula private constructor(
 
         val provisionedLoadBalancer = futureLoadBalancer.get()
         val loadBalancer = provisionedLoadBalancer.loadBalancer
+
+        val jiraNodeSecurityGroup = jiraStack.findSecurityGroup("JiraNodeSecurityGroup")
```
mzyromski-atlassian, line 250: The code that defines all the providers is quite verbose and is partially repeated in StandaloneFormula. Could we do something about it?

mzyromski-atlassian (commented):
@pczuj, awesome work! I really like the way you split the changes into separate commits. It made reviewing them much easier. 

I really like the implementation, especially the fact that the changes themselves do not break API. The only question now is: is restricting access to JPT nodes a breaking change? I'm on the fence here, as we don't have an explicit contract regarding this aspect, but I lean towards saying it is as it's an implicit requirement. If so, then I guess the corresponding in the CHANGELOG should be "Changed" instead of "Fixed" and we should produce a major release. @dagguh, what's your take on that?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
pczuj: It's a common practice for the log messages to contain issue keys formatted the Jira way. At least for me if I search in google "JPERF JPT" it gives me some results directly leading to the EAN issues, so I believe this should be fine unless the issue will need to remain closed.

I will consider creating another one as a placeholder to make it public.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
pczuj: The logic was supposed to give an opportunity to the API consumers to define many services in case one wouldn't be stable enough. In such case there can be a situation where those won't return the same IP. This situation should be treated seriously as either misconfiguration (e.g. one of the services used is broken or going though different [NIC](https://en.wikipedia.org/wiki/Network_interface_controller)) or something malicious happening (e.g. someone is doing [MITM attack](https://en.wikipedia.org/wiki/Man-in-the-middle_attack)). I can note the possibilities in the error message.

Alterntively I could make it less strict here and always take first response no matter if there were 2 or more different ones. The concern I have is that it could hide potential problem, which could have not obvious symptoms otherwise. The lib is opensource, so if the API consumer is experienced enough they should be able to easily find the error message in the code and reason based on it. Otherwise the consumer should come to us for support and it should also not be to hard to deduce what can be done do diagnose the problem further.

BTW the default service i added is one that is not documented by AWS. I'm ready to pick different one and possibly have more then one in the list for stability. Not sure which ones though.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
pczuj: Similarly to the `IllegalStateException ` triggered by `set.size != 1` I didn't think enough about this behavior to have a strong opinion. My concern with making the mechanism less strict around failures is that a silentish error like this could hide a cause while still manifesting other symptoms which are not obvious to lead towards the lack of public access. I'm seeing someone in the future failing to access JVM debug port of Jira for long time not realising to that the answer to the problem is in a single WARN line of the log (especially that there may be more WARN log lines).

The `LocalPublicIpv4Provider` is supposed to be public API and the `accessRequester` in both Server and DC formulas is overridable, so it should be rather easy to change the services used by the mechanism to the ones that the consumer has access to currently. Alternatively the initial `AccessRequester` can be disabled completely by setting it with `NoopAccessRequester`.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/UnableAccessProvider.kt
pczuj: Not sure if it wouldn't be another implementation based on the `@return` value definition I made for `AccessProvider#provideAccess()`. I mean one that would always return `true` instead of `false`, because providing an access to nothing is always certain success.

The `UnableAccessProvider` is essentially linked to that definition of the `@return`. It can't provide you any access, because it's "unable". We may change both the class name and the returned value. Let me know if you have any suggestions about the latter.

The current `@return` definition:
> @return true if it was possible to provide access, false otherwise

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
pczuj: We could, but I'm not sure if it's worth it right now.

I think that both of those formulas should be refactored and split into smaller responsibilities eventually, however it may be easier to that in bulk for all of the instructions there. I had some thoughts around it when i was experimenting around https://stash.atlassian.com/projects/HL/repos/jpt-api-reimpl-spike

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
dagguh: Even if it is a common practice, then an actual URL is a more practical practice :wink:

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/ProvisionedInfrastructure.kt
```diff
@@ -0,0 +1,34 @@
...
+import com.atlassian.performance.tools.aws.api.UnallocatedResource
+import com.atlassian.performance.tools.infrastructure.api.virtualusers.VirtualUsers
+
+class ProvisionedInfrastructure<out T : VirtualUsers>
+@Deprecated("Use ProvisionedInfrastructure.Builder instead.")
+constructor(
+    val infrastructure: Infrastructure<T>,
+    val resource: Resource
+) {
+    object Defaults {
```
dagguh, line 14: What's the upside over the inlined defaults?

dagguh, line 14: New API, is it necessary?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
```diff
@@ -3,7 +3,6 @@ package com.atlassian.performance.tools.awsinfrastructure.api.jira
 import com.amazonaws.services.cloudformation.model.Parameter
 import com.atlassian.performance.tools.aws.api.*
 import com.atlassian.performance.tools.awsinfrastructure.ApplicationStorageWrapper
-import com.atlassian.performance.tools.awsinfrastructure.InstanceAddressSelector
```
dagguh, line 6: Is `InstanceAddressSelector` used anywhere anymore? I'm not sure if it was covered by tests, but do we consciously avoid that behavior now?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/ProvisionedInfrastructure.kt
pczuj: You can't wrap (e.g. monitor) inlined defaults

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
pczuj: I believe it's not used in this repository anymore. It's not `internal` though, despite not being in the `com.atlassian.performance.tools.awsinfrastructure.api.*` package, so I'm not sure if we can remove it. Maybe we should deprecate it.

It has 2 tests, but they are not telling me exactly why it was introduced and what use case it's meant to cover. 

The functionality basically tries to connect to a given port via public IP of the instance to see if it's accessable. If it fails 4 times (1s delay) then it checks the same for private IP of that instance. It will throw an exception if neither is responding.

I found our usage of it flawed. In most of the places we were calling this mechanism with default port (ssh/22) to get an address for communicating though different port - I believe we should at least check the port which we wanted to communicate though (e.g. mysql or nfs port). This mechanism also is always run on the test coordinator machine, i.e. it's not run by VU or load balancer machine, so it made little sense to me that it determines IP though which an external machine will comunicate. I think it's better to be explicit with it.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/ProvisionedInfrastructure.kt
pczuj: It's not necessary, but it covers a lib consumer use case where it doesn't want to reconstruct a default submodule, but it wants to append it's functionality, e.g. wants to add additional resources to be released with the `Resource` class or wants to grant more permissions via `AccessProvider` or simply wants to monitor it's usages to print the diagnostics in the log or other persistance layer.

pczuj: Example: For `DataCenterFormula` you may want add more to `AccessRequester` which you can specify in the `DataCenterFormula.Builder`. You can do that by wrapping the `DataCenterFormula.Defaults.accessRequester` in `MultiAccessRequester` + your access. Otherwise you'd need to dig into the code and base your implementation on the internal implementation of the `DataCenterFormula`.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
pczuj: I'm not sure about spamming log with URLs. I tried to make the message as short as possible.

Probably we should remove it at some point. I don't know when though. Any suggestions are welcome.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
dagguh: @pbruski thoughts?

pbruski: My understanding is that we won't have public instances anymore - is that correct? If so, we don't need the address selector and can just use the private address all the time.
The class was created so that we're able to use private addresses. DCNG was only using private networking.

pbruski: Regarding the port selection, it was chosen because we use SSH to instrument the instances, so all should have that port available. That's why we also used the test coordinator.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/ProvisionedInfrastructure.kt
dagguh: Good use case, I needed it a few times in the past :+1:

dagguh: Good use case. We should log this change.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
pczuj: - [ ] Deprecate the `InstanceAddressSelector`

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/ProvisionedInfrastructure.kt
pczuj: - [ ] Add `Defauts` API to the changelog

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
pczuj: - [ ] Refactor new content of Formulas to be more readable

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/UnableAccessProvider.kt
pczuj: - [ ] Change return type of `UnableAccessProvider` to 3 value logic

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
mgrzaslewicz: I can see "JiraNodeSecurityGroup" is used 2 times in this class, maybe extract it to const?

pczuj:
- [ ] remove `TemplateBuilder#addSecurityGroupIngress`

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
pczuj: @mzyromski-atlassian, @dagguh, @mgrzaslewicz what are your thoughts about this? 

The mechanism is using a external service to check public IP address of the test executor (local machine). Current approach will throw `IllegalStateException` if this service is not working. This is not cought by anything, so the test would fail if the service would be down. 

I picked only 1 service, however this mechanism can be overwritten in the `...Formula`'s `Builder` and it's possible to set multiple services. In case 2 services would report different public IP addresses for our local machine this would also throw `IllegalStateException` resulting in test failure. If some of the services fail to report an address (e.g. they are down) only addresses reported by other services will be taken into account (or as mentioned earlier it would throw if no other services are up).

Do you think the behavior should be changed? What'd be the best logic here?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
pczuj: - [ ] Add JPERF-730 URL to the "Jira is set up" log message
- [ ] Note in the code when the JPERF-730 URL can be removed from the log message

pczuj: @dagguh, @mzyromski-atlassian, @mgrzaslewicz I'm not sure if I should add the issue key / URL to the "Jira is set up" log message at all. The rationale behind it was to mention it together with the Jira's URL, so that if user is using it they could notice that something changed.

If the issue key should be kept in this log message I'm wondering how long we should keep it? Should it be forever there or can we assume that after some time consumers will understand the new behavior? Any suggestions from you?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
pczuj: One thing I can do i to eagerly get the IP during the test static code initialisation. It would make the test fail fast in case of external service problem. Should this be done?
```
val accessRequester: AccessRequester = ForIpAccessRequester(LocalPublicIpv4Provider().get().let { ip -> { ip }})
```
instead of 
```
val accessRequester: AccessRequester = ForIpAccessRequester(LocalPublicIpv4Provider())
```

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/UnableAccessProvider.kt
pczuj: A while ago I discussed the topic of values returned by `AccessProvider`/`AccessRequester` with @dagguh. We agreed on changing it to 3 value logic to model service-like response (like http 200/400/500 response codes), however now when I'm trying to implement it I again think it should be left as simple as it is. I don't want to model exact behavior of the implementation underneath this interface. The caller should not be interested in recovering from any failures - it's up to the implementation to provide as reliable logic as possible. The problem is that we are not injecting dependencies (i.a. requester factory) into the domain level logic. For it to be done properly this code would need to be restructured from ground up and I believe we don't have time for that.

I'm fine with changing the name to `NoAccessProvider` and making this class return `true` to indicate that the task of not granting any access was successful. I could then also change the definition of returned value to `true if access grant was successful, false otherwise`, as it may be more intuitive than `true if it was possible to grant access, false otherwise`. 

@dagguh, @mzyromski-atlassian, @mgrzaslewicz what do you think about it?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
mzyromski-atlassian: I'm on the fence here. I think we can even drop the issue link altogether assuming we'll have a proper, detailed description of the change in the changelog. I don't have a strong opinion, so whatever you decide @pczuj will be fine for me.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
mzyromski-atlassian: OK, let's keep it as is.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
mzyromski-atlassian: Given that the service is actually provided by AWS, I guess we can expect it to be quite stable. I agree that failing fast might be a better approach than being surprised later. The eager IP fetch sounds like a plan too.

pczuj: - [ ] Resolve the public IP eagerly
- [ ] Explain what can be done by the API consumer in the error messages

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/UnableAccessProvider.kt
mzyromski-atlassian: +1 for renaming to `NoAccessProvider` and following the `true if it was possible to grant access, false otherwise` convention.

mzyromski-atlassian:
@pczuj, I think I answered all your questions. In general, I like the code as it is now. If you can implement the small fixes discussed in the threads, that's even better.

The open question is still: is this a breaking change that will required major release of the library? And here, again, I'm on the fence. If you don't feel it is, then let's consider it as "Fixed" (instead of "Changed") type of change and release as a minor version.

mzyromski-atlassian (dismissed):

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/UnableAccessProvider.kt
pczuj: - [ ] Change the `AccessProvider` return value doc
- [ ] Rename `UnableAccessProvider` to `NoAccessProvider`

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/SecurityGroupIngressAccessProvider.kt
```diff
@@ -0,0 +1,65 @@
...
+        AuthorizeSecurityGroupIngressRequest()
+            .withGroupId(securityGroup.groupId)
+            .withIpPermissions(ipPermissions)
+    ).sdkHttpMetadata
+        .httpStatusCode / 100 == 2
+
+    class Builder(
+        private val ec2: AmazonEC2,
+        private val securityGroup: SecurityGroup,
+        private val portRange: IntRange
```
mgrzaslewicz, line 52: I can see the only values passed here are single ports across the whole project. Is range needed here?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/AccessProvider.kt
```diff
@@ -0,0 +1,22 @@
...
+ * Used to grant access to a network resource, e.g. web application.
+ * Introduced to allow restriction of anonymous access.
+ */
+interface AccessProvider {
+
+    /**
+     * @param cidr [Classless Inter-Domain Routing](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing) format of IP and network mask
+     * @return true if it was possible to provide access, false otherwise
+     */
+    fun provideAccess(cidr: String): Boolean
```
mgrzaslewicz, line 15: how about longer variable name? Without reading docs it's unclear

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
```diff
@@ -40,6 +43,7 @@ class InfrastructureFormula<out T : VirtualUsers> private constructor(
 ) {
     private val logger: Logger = LogManager.getLogger(this::class.java)
 
+    @Deprecated("Use InfrastructureFormula.Builder instead.")
     constructor(
```
mgrzaslewicz, line 47: AwsDatasetModification is using it, consider removing the usage

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
mgrzaslewicz: Spending a lot time lately with JPT I'd be frustrated not being able to log into the db or Jira host after some failure.
How often does it fail?

mgrzaslewicz: > @mzyromski-atlassian, @dagguh, @mgrzaslewicz what are your thoughts about this?
> 
> The mechanism is using a external service to check public IP address of the test executor (local machine). Current approach will throw `IllegalStateException` if this service is not working. This is not cought by anything, so the test would fail if the service would be down.
> 
> I picked only 1 service, however this mechanism can be overwritten in the `...Formula`'s `Builder` and it's possible to set multiple services. In case 2 services would report different public IP addresses for our local machine this would also throw `IllegalStateException` resulting in test failure. If some of the services fail to report an address (e.g. they are down) only addresses reported by other services will be taken into account (or as mentioned earlier it would throw if no other services are up).
> 
> Do you think the behavior should be changed? What'd be the best logic here?

Spending a lot time lately with JPT I'd be frustrated not being able to log into the db or Jira host after some failure.
How often does it fail?

mgrzaslewicz:
After this change - will is still be possible to from atlassian network to ssh to the machine as usual?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/SecurityGroupIngressAccessProvider.kt
pczuj: This API was designed to match the underlying mechanism (Security Group Ingress) and it should stay like that, because this models that mechanism. Changing it to single port would be unnecessary limitation.

If we want to achieve simplification of the API we can do another implementation that has only a single port - this one could use the `SecurityGroupIngressAccessProvider`underneath.

I was also experimenting with something different. I believe `DataCenterFormula` should delegate more repsponsibilities outside to constructor injected dependencies. Essentially no dependency objects should be created as part of provisioning itself. We could define "AccessProviderFactory" with API like `fun createAccessProvider(port: Int, securityGroup: SecurityGroup): AccessProvider` (actually that would be `SecurityGroupPortAccessProviderFactory`) and use that instead of calling constructor of specific class. It could provide us an ability to further observe what the formula is doing and API consumers could also override this functionality in different ways. The problem with thts approach is that it fully exposes the internal dependencies of the class - which should be fine if we treat every implementation dependencies as final, however this is not how the code in this repo was made so far.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/AccessProvider.kt
pczuj: I'm not sure what do you mean. Could you clarify?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/InfrastructureFormula.kt
pczuj: - [ ] Convert `AwsDatasetModification`'s usage of `InfrastructureFormula` to it's `Builder`

pczuj:
@mgrzaslewicz yes, the ssh port configuration is unchanged. The modification is only about potentially insecure application ports. We are mostly interested in http, db and jvm specific ports.

There was a suggestion from Maciej during our zoom discussion to change the definition of ssh port access to also use the new `AccessProvider`, so that we don't use 2 mechanisms (i.e. drop the ingress rule definition from static cloudformation template file). Do you think that could make it more clear if the ssh access is still present? Maybe it should also be clarified in the CHANGELOG?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
pczuj: > How often does it fail?
I don't expect that to ever happen, however you never know... with this we are making the whole code dependant on an external service functionality. Sure it can be overwritten, but we could end up in a situation in the future that for "old" versions of aws-infrastructure we would always need to overwrite it, because the service we are using no longer exists. We could quickfix it though.

I believe I will pick some backup services for this.

- [ ] Add backup public ip services
- [ ] Ad unit tests for the public ip service fallback

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/AccessProvider.kt
mgrzaslewicz: I mean the 'cidr' is not obvious without reading the doc for method.
Maybe something like 'ipWithMask' as parameter name? + example in comment not to reach wiki to check it
* @param cidr [Classless Inter-Domain Routing](https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing) format of IP and network mask
* e.g. 198.51.100.14/24

mgrzaslewicz:
> @mgrzaslewicz yes, the ssh port configuration is unchanged. The modification is only about potentially insecure application ports. We are mostly interested in http, db and jvm specific ports.
> 
> There was a suggestion from Maciej during our zoom discussion to change the definition of ssh port access to also use the new `AccessProvider`, so that we don't use 2 mechanisms (i.e. drop the ingress rule definition from static cloudformation template file). Do you think that could make it more clear if the ssh access is still present? Maybe it should also be clarified in the CHANGELOG?

clarification in changelog would be fine. However ssh'ing to created hosts should be made much easier than it currently is and that's a separate topic which I raised here: https://ecosystem.atlassian.net/browse/JPERF-761

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/AccessProvider.kt
pczuj: I used the terminology that AWS API uses. Not sure if I should interpret it on our level.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
```diff
@@ -0,0 +1,61 @@
+package com.atlassian.performance.tools.awsinfrastructure.api.network.access
+
+import com.atlassian.performance.tools.awsinfrastructure.Ipv4Validator
+import com.atlassian.performance.tools.awsinfrastructure.VerboseUrlReader
+import java.net.URL
+import java.util.function.Supplier
+
+class LocalPublicIpv4Provider private constructor(
```
dagguh, line 8: What's local about it? Doesn't it provide IP as seen by the `servicesToQuery`?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
```diff
@@ -0,0 +1,61 @@
...
+    private val servicesToQuery: List<URL>,
+    private val urlReader: (serviceUrl: URL) -> String?,
+    private val ipValidator: (ip: String) -> Boolean
+) : () -> String, Supplier<String> {
+    object Defaults {
+        val servicesToQuery: List<URL> = listOf(
+            URL("https://checkip.amazonaws.com")
```
dagguh, line 15: Do you rely on the DNS lookup inside `URL.equals`? If not, we should use `URI` instead.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
```diff
@@ -0,0 +1,61 @@
...
+        )
+        val urlReader: (serviceUrl: URL) -> String? = VerboseUrlReader()
+        val ipValidator: (ip: String) -> Boolean = Ipv4Validator()
+    }
+
+    constructor() : this(
```
dagguh, line 21: We've got a builder with defaults already. What's the use case for the extra constructor? IDE autocompletion?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
dagguh: > backup public ip services

YAGNI. Add backup only when the default service is actually flaky.

dagguh: > Resolve the public IP eagerly

Don't do any work during initialization. Especially if the effect of the default work can be overridden via builder.

dagguh: > should we fail or just warn and go on? 

That's a decision outside of `LocalPublicIpv4Provider`.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
pczuj: The "Local" part of name of this class was to indicate what instance this public IP is resolved for. In theory there could be a public IP address resolver for remote instances.

Do you have any other name suggestions for this class?

Side topic: I was thinking this weekend about class naming and came into a conclusion that implementations shouldn't have names. What we are naming is the contract which should be represented by an interface. IMHO it would be better to name implementations as `<interface name>V1`,`<inteface name>V2` and so on. I don't think we should start naming classes like that right now, but maybe this is something to brainstorm and consider for other projects.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
pczuj: I don't know about relying on it. Would you be able to explain or share link to explanation why this matters?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
pczuj: I used it to avoid need of writing `LocalPublicIpv4Provider.Builder().build()` and instead just be able to `LocalPublicIpv4Provider()`. Do you think it's not worth it and I should remove this constructor and always use the `Builder`?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
dagguh: [`URL.equals` Javadoc](https://docs.oracle.com/javase/9/docs/api/java/net/URL.html#equals-java.lang.Object-):
>  Two hosts are considered equivalent if both host names can be resolved into the same IP addresses; else if either host name can't be resolved, the host names must be equal without regard to case; or both host names equal to null.
> Since hosts comparison requires name resolution, this operation is a blocking operation.
> Note: The defined behavior for equals is known to be inconsistent with virtual hosting in HTTP.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
dagguh: Aaah, I thought it's a "local provider of public ipv4", but it's a "provider of local public ipv4". All good :+1:

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
dagguh: It's ok. Empathizing with callers is a good thing :slightly_smiling_face:

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
dagguh: > implementations shouldn't have names

Obviously current interfaces wouldn't fit: `SupplierOfStringV1`, `SupplierOfStringV2`, `SupplierOfStringV1329`. We'd have to capture all behaviors and aspects in the interface itself (e.g. local instance + public ipv4 + services to query + ip validation). I met an argument that variable names shouldn't be necessary, because the type should be specific enough to have only one possible meaning. These are interesting ideas to explore and experiment with, but I wouldn't bake them into API just yet.

pczuj: > Obviously current interfaces wouldn't fit

Yup, I didn't create a dedicated interface for it, because of laziness. Also it felt unnecessary.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
pczuj: If I shouldn't use `URL` here then I'd rather use `String` for the service addresses. With URI I will have problem around spliting type conversion responsibilities.

pczuj:
@dagguh do you think this change could be released as minor version bump of aws-infrastructure?

I'm not sure if we ever made the API out of public accessability of JPT instances ports. Some consumers may rely on it though.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/LocalPublicIpv4Provider.kt
pczuj: > YAGNI. Add backup only when the default service is actually flaky.

I disagree. It's part of the solution quality to not rely on a single service by default. I'm still adding additional services, because it's low cost.

> Don't do any work during initialization. Especially if the effect of the default work can be overridden via builder.

I agree. Will skip this.

> That's a decision outside of `LocalPublicIpv4Provider`.

I'd like to avoid creating API out of exceptions. I will still leave the code as it is around it, because if someone would want to have it differently it's possible to overwrite this dependency. I'm only adding more verbose explanation of what can be done in case of given exception happening.

dagguh:
Current situation: public API didn't contain a behavioral contract about port access (neither via explicit JavaDoc nor via implicit class/method names).

Possible next steps:
* release as fix, impacting use cases of some clients
* add as behavioral contract, then break it, slowing down the adoption of security fix

I would not promote the current hidden behavior as API. Because clients shouldn't hack their use cases around this stuff, they should ask for APIs that can reliably support their use cases.

pczuj:
I pushed implementation of PR suggestions as separate commits. I'm going to squash them to previous commits, however for transparency I wanted to push them in this way first, so you could review commit by commit what changed.

pczuj:
@dagguh thanks for confirmation. Let's treat it as a fix with new API.

I'm gonna use the [japicmp](https://siom79.github.io/japicmp/index.html) to see if I didn't break any code contract.

mzyromski-atlassian (approved):

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
pczuj: Out of all suggestions this one I didn't manage to implement.

My idea is to break down the responsibilities of `DataCenterFormula` into it's smaller dependencies (interfaces) which could be overwritten, but would have their default implementations (classes). We could split things like:
 * Provisioner
 * Instance Identifier
 * Ssh/IP Resolver
 * Access Definition (this one would be the one around my changes)

I don't have time to finish it and the refactoring should be possible later on, so I'd like to proceed without it.

CHANGELOG.md
```diff
@@ -23,6 +23,24 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-2.23.1...master
 
+### Fixed
+- Restrict access to provisioned AWS instances to only the necessary connections. Resolve [JPERF-730].
+
+### Added
+- Add the ability to dynamically share permissions between resources provisioned by various Formulas. Unblock [JPERF-730].
+- Add builders for `ProvisionedJira`, `ProvisionedLoadBalancer`, `ProvisionedVirtualUsers` and `ProvisionedInfrastructure`.
+- Add `Defaults` child object to `ProvisionedJira`, `ProvisionedLoadBalancer`, `ProvisionedVirtualUsers`, `ProvisionedInfrastructure`, `StandaloneFormula` and `DataCenterFormula`.
```
pczuj, line 32: @dagguh this is application of your suggestion. Is this CHANGELOG entry about `Defaults` good enough or would you suggest framing it differently?

mzyromski-atlassian:
> @mzyromski-atlassian The only thing I didn't implement out of suggestions is [the simplification of the code added to `JiraFormula` implementations (DC/Standalone)](https://github.com/atlassian/aws-infrastructure/pull/116#discussion_r757087907). I have an idea how this could be done, but it would take me some time to delegate part of responsibilities into dependencies injected into the formulas. I'm worried that I don't have time for that right now. Is that fine?

Yes, I'm fine with that. Thanks.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/access/AccessProvider.kt
pczuj: @mzyromski-atlassian what do you think about this suggestion?

mzyromski-atlassian: I think CIDR should be clear for a person knowing the context (AWS).

src/main/resources/aws/2-nodes-dc.yaml
```diff
@@ -127,75 +123,57 @@ Resources:
...
       VpcId:
         Ref: Vpc
-      GroupDescription: Enables SSH access
+      GroupDescription: For SSH enabled instances
       SecurityGroupIngress:
         -
           IpProtocol: tcp
           FromPort: 22
           ToPort: 22
           CidrIp: 0.0.0.0/0
```
pczuj, line 132: @dagguh I recalled that during our video call in the past you suggested that SSH access should also be granted by `AccessProvider` similarly to other port access. That would be to avoid complexity of using 2 different mechanisms.

pczuj: I didn't implement this suggestion and I'm thinking if I could skip it, because there is pressure to merge this PR

CHANGELOG.md
dagguh: Looks good

src/main/resources/aws/2-nodes-dc.yaml
dagguh: Yep, the scope is quite big. Raised https://ecosystem.atlassian.net/browse/JPERF-763

pczuj:
Rebased the commits interactively (`git rebase -i origin/master`) like this:
```
pick 251f5c2 JPERF-730: Set memory arguments to avoid java.lang.OutOfMemoryError: Metaspace
pick 925c9b2 JPERF-730: Bump AWS SDK and httpclient dependency versions to make it compatible with recent versions of aws-resources, which contains new API necessery for JPERF-$
pick a3bfe6c Deprecate InfrastructureFormula constructor, to promote usage of it's builder, which is more flexible for future API evolution
f f097f80 JPERF-730: Remove deprecated usage of `InfrastructureFormula` constructor
pick a75d281 JPERF-730: Extract "Provisioned" classes into their own files and add Builders for them, so we can extend their API easily without breaking it
pick 29631aa Add diagnostics to know which node failed to start as part of DataCenterFormula
pick 0337d21 JPERF-730: Refactor "Formulas", so that Standalone matches DC and it's more explicit which IP addresses are used for what
f ab47f3d JPERF-730: Delete InstanceAddressSelector
pick b8db7d8 JPERF-730: Make TemplateBuilder more versitile around adding SecurityGroupIngress rules, correct it's description modification mechanism and add diagnostics around$
pick 92064a5 JPERF-730: Restrict access to AWS instances provisioned by JPT only to the necessery connections
f 7d2ff06 JPERF-730: Add mention about Defaults to the CHANGELOG
f 1e723c3 JPERF-730: Drop the change of "Jira is set up" log message noting recent permission changes to avoid need of it's later maintenance
f a140a58 JPERF-730: Change URL usages to URI usages inside LocalPublicIpv4Provider to avoid collections of URL.equals in URL collections
f 066b67e JPERF-730: Explain what can be done by the API consumer in case of public IP provider errors
f 69a74cc JPERF-730: Add more public IP services to query by default, so we can avoid relying on a single one in case of it's maintenance or failure
f 4031b81 JPERF-730: Explain what "Local" means in the context of `LocalPublicIpv4Provider` class
f 97dcdd3 JPERF-730: Fail service URI read with null instead of rethrow
f 0b6b63f JPERF-730: Change the `AccessRequester`/`AccessProvider` return value contract to simplify it and modify implementations accordingly
f ef0b568 JPERF-730: Remove unused TemplateBuilder code
f 8920f0a JPERF-730: Add unit tests for `LocalPublicIpv4Provider`
```

pczuj:
I expect the current state of thie PR to be final. Let me know if I overlooked some crucial feedback/suggestion which should be applied.

Next I'm gonna run some manual tests and if those will not reveal any problems then my intent is to merge this. 

"Some manual tests":
 * [japicmp](https://siom79.github.io/japicmp/index.html)
 * Running the lib with our internal tests

pczuj:
This is the output from japicmp comparing my locally built aws-infrastructure from this branch with version 2.23.1 (without any exclusion rules):
```
Status | Fully Qualified Name
-- | --
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.AmiNameResolver
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.AmiNameResolver$AwsAndAmi
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.AmiNameResolver$Companion
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.aws.AwsCli
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.AwsDatasetModification
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.AwsDatasetModification$Builder
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.CustomDatasetSource
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.CustomDatasetSource$Builder
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.CustomDatasetSource$FileNames
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.dataset.DatasetHost
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.dataset.S3DatasetPackage
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.dataset.SshMysqlDatasetPublication
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.DatasetCatalogue
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.hardware.C4EightExtraLargeElastic
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.hardware.C5NineExtraLargeEphemeral
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.hardware.Computer
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.hardware.EbsEc2Instance
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.hardware.M4ExtraLargeElastic
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.hardware.M5ExtraLargeEphemeral
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.hardware.Volume
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.Infrastructure
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.InfrastructureFormula
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.InfrastructureFormula$Builder
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.InfrastructureKt
MODIFIED | com.atlassian.performance.tools.awsinfrastructure.api.jira.DataCenterFormula
MODIFIED | com.atlassian.performance.tools.awsinfrastructure.api.jira.DataCenterFormula$Builder
NEW | com.atlassian.performance.tools.awsinfrastructure.api.jira.DataCenterFormula$Defaults
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.jira.Jira
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.jira.JiraFormula
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.jira.JiraSoftwareDevDistribution
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.jira.JiraSoftwareInternalDistribution
MODIFIED | com.atlassian.performance.tools.awsinfrastructure.api.jira.ProvisionedJira
NEW | com.atlassian.performance.tools.awsinfrastructure.api.jira.ProvisionedJira$Builder
NEW | com.atlassian.performance.tools.awsinfrastructure.api.jira.ProvisionedJira$Defaults
MODIFIED | com.atlassian.performance.tools.awsinfrastructure.api.jira.StandaloneFormula
MODIFIED | com.atlassian.performance.tools.awsinfrastructure.api.jira.StandaloneFormula$Builder
NEW | com.atlassian.performance.tools.awsinfrastructure.api.jira.StandaloneFormula$Defaults
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.jira.StartedNode
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.jira.UriJiraFormula
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.loadbalancer.ApacheEc2LoadBalancerFormula
MODIFIED (!) | com.atlassian.performance.tools.awsinfrastructure.api.loadbalancer.ApacheProxyLoadBalancer
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.loadbalancer.ElasticLoadBalancerFormula
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.loadbalancer.LoadBalancerFormula
MODIFIED | com.atlassian.performance.tools.awsinfrastructure.api.loadbalancer.ProvisionedLoadBalancer
NEW | com.atlassian.performance.tools.awsinfrastructure.api.loadbalancer.ProvisionedLoadBalancer$Builder
NEW | com.atlassian.performance.tools.awsinfrastructure.api.loadbalancer.ProvisionedLoadBalancer$Defaults
NEW | com.atlassian.performance.tools.awsinfrastructure.api.network.access.AccessProvider
NEW | com.atlassian.performance.tools.awsinfrastructure.api.network.access.AccessRequester
NEW | com.atlassian.performance.tools.awsinfrastructure.api.network.access.ForCidrAccessRequester
NEW | com.atlassian.performance.tools.awsinfrastructure.api.network.access.ForIpAccessRequester
NEW | com.atlassian.performance.tools.awsinfrastructure.api.network.access.ForSecurityGroupAccessRequester
NEW | com.atlassian.performance.tools.awsinfrastructure.api.network.access.LocalPublicIpv4Provider
NEW | com.atlassian.performance.tools.awsinfrastructure.api.network.access.LocalPublicIpv4Provider$Builder
NEW | com.atlassian.performance.tools.awsinfrastructure.api.network.access.LocalPublicIpv4Provider$Defaults
NEW | com.atlassian.performance.tools.awsinfrastructure.api.network.access.MultiAccessProvider
NEW | com.atlassian.performance.tools.awsinfrastructure.api.network.access.MultiAccessRequester
NEW | com.atlassian.performance.tools.awsinfrastructure.api.network.access.NoAccessProvider
NEW | com.atlassian.performance.tools.awsinfrastructure.api.network.access.NoAccessRequester
NEW | com.atlassian.performance.tools.awsinfrastructure.api.network.access.SecurityGroupIngressAccessProvider
NEW | com.atlassian.performance.tools.awsinfrastructure.api.network.access.SecurityGroupIngressAccessProvider$Builder
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.network.Network
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.network.NetworkFormula
MODIFIED | com.atlassian.performance.tools.awsinfrastructure.api.ProvisionedInfrastructure
NEW | com.atlassian.performance.tools.awsinfrastructure.api.ProvisionedInfrastructure$Builder
NEW | com.atlassian.performance.tools.awsinfrastructure.api.ProvisionedInfrastructure$Defaults
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.RemoteLocation
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.storage.ApplicationStorage
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.storage.ApplicationStorage$DefaultImpls
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.storage.JiraServiceDeskStorage
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.storage.JiraSoftwareStorage
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.storage.S3Artifact
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.storage.S3TarGzDistribution
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.TargetingVirtualUserOptions
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.virtualusers.AbsentVirtualUsers
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.virtualusers.AbsentVirtualUsersFormula
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.virtualusers.Ec2VirtualUsersFormula
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.virtualusers.Ec2VirtualUsersFormula$Builder
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.virtualusers.MulticastVirtualUsersFormula
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.virtualusers.MulticastVirtualUsersFormula$Builder
MODIFIED | com.atlassian.performance.tools.awsinfrastructure.api.virtualusers.ProvisionedVirtualUsers
NEW | com.atlassian.performance.tools.awsinfrastructure.api.virtualusers.ProvisionedVirtualUsers$Builder
NEW | com.atlassian.performance.tools.awsinfrastructure.api.virtualusers.ProvisionedVirtualUsers$Defaults
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.virtualusers.StackVirtualUsersFormula
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.virtualusers.StackVirtualUsersFormula$Builder
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.api.virtualusers.VirtualUsersFormula
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.ApplicationStorageWrapper
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.AvaliabilityZonePickerKt
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.hardware.EphemeralStorage
REMOVED (!) | com.atlassian.performance.tools.awsinfrastructure.InstanceAddressSelector
REMOVED (!) | com.atlassian.performance.tools.awsinfrastructure.InstanceAddressSelector$Companion
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.InstanceFilters
NEW | com.atlassian.performance.tools.awsinfrastructure.Ipv4Validator
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.jira.DataCenterNodeFormula
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.jira.DataCenterNodeFormula$provision$2
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.jira.DiagnosableNodeFormula
NEW | com.atlassian.performance.tools.awsinfrastructure.jira.DiagnosableStoppedNode
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.jira.home.SharedHomeFormula
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.jira.NodeFormula
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.jira.StandaloneNodeFormula
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.jira.StandaloneStoppedNode
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.jira.StoppedNode
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.loadbalancer.ElasticLoadBalancer
MODIFIED (!) | com.atlassian.performance.tools.awsinfrastructure.TemplateBuilder
NEW | com.atlassian.performance.tools.awsinfrastructure.VerboseUriReader
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.virtualusers.S3ResultsTransport
UNCHANGED | com.atlassian.performance.tools.awsinfrastructure.virtualusers.UbuntuVirtualUsersRuntime

Binary incompatible changes are marked with (!) while source incompatible changes are marked with (*).
```

In such config it suggest major version update, however I believe the breaking changes are not in our API.

> MODIFIED (!) | com.atlassian.performance.tools.awsinfrastructure.api.loadbalancer.ApacheProxyLoadBalancer

`ApacheProxyLoadBalancer` is an `internal` class however it's inside `...api...` package. I changed it's constructor. Should we treat this as breaking?

> REMOVED (!) | com.atlassian.performance.tools.awsinfrastructure.InstanceAddressSelector
> REMOVED (!) | com.atlassian.performance.tools.awsinfrastructure.InstanceAddressSelector$Companion
> MODIFIED (!) | com.atlassian.performance.tools.awsinfrastructure.TemplateBuilder

All of those are `internal` classes outside of `...api...` package.

When I applied includes/excludes to the plugin in such way:
```
<includes>
    <include>com.atlassian.performance.tools.awsinfrastructure.api.*</include>
</includes>
<excludes>
    <exclude>com.atlassian.performance.tools.awsinfrastructure.api.loadbalancer.ApacheProxyLoadBalancer</exclude>
</excludes>
<includeExclusively>true</includeExclusively>
```

The result suggest minor version bump.

dagguh:
Kotlin's `internal` is not recognized by Java, so they become API, when in `api` package.

pczuj:
I'm gonna add `Builder` to the `ApacheProxyLoadBalancer`, move it to separate file and remove `internal` from it as it's misleading (Could this removal be API breaking?). Also I will deprecate constructors.

pczuj:
Again added the change as a separate commit for PR visibility, but I'm also rebasing it now to squash it into existing ones for clear git history.

pczuj:
After those changes japicmp with the configuration is suggesting minor version update 👌
```
<includes>
    <include>com.atlassian.performance.tools.awsinfrastructure.api.*</include>
</includes>
<includeExclusively>true</includeExclusively>
```

dagguh:
> Could this removal be API breaking?

When in doubt, check the bytecode (e.g. `Show Kotlin Bytecode` action in IntelliJ).

Diffing the whole bytecode shows only a change in a `@Lkotlin/Metadata;`.

The class is actually public with `internal` or without it:

```java
// ================com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/ApacheProxyLoadBalancer.class =================
// class version 52.0 (52)
// access flags 0x31
public final class com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/ApacheProxyLoadBalancer implements com/atlassian/performance/tools/infrastructure/api/loadbalancer/LoadBalancer {
```

The Kotlin annotation is respected by JetBrains IDE, but I doubt it works anywhere else. There's a big [debate around package-private in Kotlin](https://youtrack.jetbrains.com/issue/KT-29227).

mzyromski-atlassian (approved):

src/main/resources/aws/load-balancer.yaml
```diff
@@ -39,15 +39,10 @@ Resources:
             -
               Name: "CookieName"
               Value: "JSESSIONID"
-  HttpSecurityGroup:
+      Scheme: internal
```
pczuj, line 42: My final tests revealed that I forgot to remove this one line after changing the approach from fully internal communication to dynamicly assigned access.

With this `Scheme: internal` the load balancer is only accessable from internal network, which was intended when I was passing it's private IP to the VUs.

Problematic thing is that we don't have integration tests in this repo for this load balancer. We are using it internally though, so running those with bumped aws-infrastrusture dependency SNAPSHOT failed.

Unfortunately that means I will need to add at least one more commit and @mzyromski-atlassian is on holidays, so he won't be able to appove. I'm counting on other reviewers.

pczuj:
squashing the last commit

pczuj:
Ok, this should be the final version. I'm waiting for CI and for eventual approve then would like to merge.

pbruski (approved):

gregrow (approved):

mzyromski-atlassian (approved):

PR #117, opened 2021-11-24
JPERF-273 AWS DC with hooks

dagguh:


pczuj:
@dagguh are you still planning to work on this or would you like to hand it over?

dagguh:
I wanted to finish it, but when I heard rumors about outsourcing JPT, I got demotivated.

pczuj:
In Jira DC we won't stop using JPT unless there will be a tool with similar functional and **non-functional** requirements (or I'm out and someone else responsible for our performance testing would accept lack of some non-functional requirements and make our tests to be replaced by a solution with less maintainability/extendability).

I personally think the improvement in infrastructure setup flexibility is needed if we want to invest into our tests and it seems this is one of the priorities of the team. I thought this PR could be looked at again, though I'm not commiting that I will be able to help in any way around it yet.

I will be honest that the way I'm used to modeling the systems is a bit in contrast to how those changes are made and actually how a lot of JPT is already written. I thought that I can use this opportunity to learn or possibly fix the implementation model.

dagguh:
> infrastructure setup flexibility is needed if we want to invest into our tests and it seems this is one of the priorities of the team

I see. I'll resurrect this. Note that I can only spend 20% of time, so if we need to speed up, we can pair up for partial handover.

dagguh:
PS. This is widening the range of hypotheses that JPT can test.
However, I'd also consider deepening the insights we're getting from each test. JPT is good for detecting regressions, but not good enough for fixing them. "Rollback the last change" is often not enough when we want to move forward with Jira features. We could do automatic analysis of frontend waterfalls, backend profiler flamecharts. Finding critical paths, tracing requests and threads, even using linear regression models. Just a hint for unlocking true potential of JPT.

pczuj:
However I've got impression that right now our tests are not really representing scenarios close enough to real product usage. TBH I'm not even sure if it's practically possible to create a regression detection pipeline that will cover most of the cases that our users rely on. Cloud is better here.

The other capability that JPT grants is an environment where devs can experiment. Team expects us to provide them the tooling for proving or disproving theories that they are creating based on their understanding of the product code. I feel like I fail miserably when they ask me how to test something and I need to tell them that we don't have an functionality out of the box. Test customisation in JPT is currently not easy enough.

PR #118, opened 2021-11-24
Hl 2178 adjust to infrastructure changes

mgrzaslewicz:


pczuj:
Mind that changes from #116 are conficting with those. Rebase should be trivial though.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
```diff
@@ -67,7 +68,8 @@ class DataCenterFormula private constructor(
         application: com.atlassian.performance.tools.awsinfrastructure.api.storage.ApplicationStorage,
         jiraHomeSource: JiraHomeSource,
         database: Database,
-        computer: Computer
+        computer: Computer,
+        adminPasswordPlainText: String
```
pczuj, line 72: I believe you shouldn't change existing deprecated constructors. For new fields required by the main constructor you should provide a default there.

In #116 I'm introducing new pattern of `Default` object that can be used to share defaults between deprecated constructors and builders, so that those can be defined in one place. I think you could reuse this pattern unless other JPT contributors have anything against.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneStoppedNode.kt
```diff
@@ -97,11 +98,16 @@ internal data class StandaloneStoppedNode(
         return ssh.execute("cat $unpackedProduct/work/catalina.pid").output.trim().toInt()
     }
 
+    /**
+     * Possible side effect: when wrong password provided increases cwd_user.invalidPasswordAttempts and that might result in
```
pczuj, line 102: Can this happen via this automation for some reason?

build.gradle.kts
```diff
@@ -41,11 +41,11 @@ configurations.all {
 }
 
 dependencies {
-    api("com.atlassian.performance.tools:infrastructure:[4.13.0,5.0.0)")
+    api("com.atlassian.performance.tools:infrastructure:4.18.1-SNAPSHOT")
```
pczuj, line 44: - [ ] Remember to change this to non-SNAPSHOT before merging

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
mgrzaslewicz: How long do we want to keep deprecated constructor? Maybe we could just remove those

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneStoppedNode.kt
mgrzaslewicz: What do you mean by automation? Comment explains it does happen when using rest calls with wrong password provided

pczuj: From what I understand the described behavior is related to the Jira produce and I understand that the problem around captcha can happen when an automation is using the wrong password. 

My question is: Can this wrong password happen in our solution? If yes than maybe we could detect in the code and inform with verbose exception message.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
pczuj: There are 2 approaches. I'm not sure which JPT is using:
 * We could remove them at any point and bump major version when releasing the code.
 * We could leave them forever, as they provide value and we could live with them. Major versions don't need to break compatibility.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneStoppedNode.kt
mgrzaslewicz: I think this scenario is 99% unlikely to appear again. We control password being set in db and used for virtual user.
I've found no other cases of hardcoded usage of credentials.
And for that 1% of uncertanity there is a comment, detecting counter of invalidPasswordAttempts before login would be complicated to implement

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
dagguh: We should remove them and do a major-bump when the pain of maintaining too many deprecations is bigger than the major release cascade.We should re

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
```diff
@@ -427,6 +435,8 @@ class DataCenterFormula private constructor(
 
         fun databaseVolume(databaseVolume: Volume): Builder = apply { this.databaseVolume = databaseVolume }
 
+        fun adminPasswordPlainText(adminPasswordPlainText: String): Builder = apply { this.adminPasswordPlainText = adminPasswordPlainText }
```
dagguh, line 438: New API needs a callout in the changelog

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/StandaloneFormula.kt
```diff
@@ -227,14 +231,14 @@ class StandaloneFormula private constructor(
         val setupDatabase = executor.submitWithLogContext("database") {
             databaseSsh.newConnection().use {
                 databaseComputer.setUp(it)
-                logger.info("Setting up database...")
+                logger.info("Setting up database with ip ${databaseSshIp}...")
```
dagguh, line 234: Why log IP in this specific case? Should all the other provisioning logs include IPs?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneStoppedNode.kt
dagguh: This is a good source code comment. In case where you use a wrong password and then somehow reuse that node with a correct password, VUs could fail due to captcha. I dunno how to construct such a use case, but apparently it already has bitten @mgrzaslewicz so documenting this "gotcha" is a plus.

dagguh (changes requested):

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/StandaloneFormula.kt
mgrzaslewicz: it helped a lot during analyzing problems allowing going straight to DB host.
I've also raised https://ecosystem.atlassian.net/browse/JPERF-761 - I think it should be much easier to ssh to any provisioned host

dagguh: That feature is already done, see the Jira comment.

mgrzaslewicz: I've checked it and indeed it is. I've removed the ip from logs

mzyromski-atlassian (approved):

dagguh (approved):

PR #119, opened 2021-12-20
Update log4j to 2.17 to address CVE-2021-45105.

jagveer-loky:
https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-45105

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -69,7 +69,7 @@ dependencies {
 fun log4j(
     vararg modules: String
 ): List<String> = modules.map { module ->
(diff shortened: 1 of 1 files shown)


mzyromski-atlassian:
@jagveer-loky, thanks for your PR, I appreciate it. There are some missing bits here, e.g. lock files need to be modified and also there are some problems with CI. At the same time we (Atlassian) started working on addressing the log4j issue in all JPT components, hence we should have a similar PR soon that will in a way supersede this one. First, we need to take a look at the CI problem though.

PR #125, opened 2022-01-26
JPERF-792: Fix Jira Data Center testing for pre version 8.9.0

pczuj:
This was missed when reporting and fixing JPERF-790.

For Jira pre 8.9.0 if the instance has no access to its own HTTP the dashboard view may freeze (in our case it was after log in, however it may be related to dataset config).

This access to self is described as required in the setup documentation https://confluence.atlassian.com/jirakb/configure-linux-firewall-for-jira-applications-741933610.html and it was missed in implementation of aws-infrastructure 2.24.0

> 4 - Allowing connections to JIRA from itself (to ensure you don't run into problems with [gadget titles showing as __MSG_gadget](https://confluence.atlassian.com/jirakb/fix-gadget-titles-showing-as-__msg_gadget-in-jira-server-813697086.html))
> ```iptables -t nat -I OUTPUT -p tcp -o lo --dport 80 -j REDIRECT --to-ports 8080```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 
 ## [Unreleased]
 
+### Fixed
+- Grant load balancer access to every Jira node provisioned by `DataCenterFormula`. Fix [JPERF-792].
+
+[JPERF-792]: https://ecosystem.atlassian.net/browse/JPERF-792
+
 ## [2.25.2] - 2022-01-21
 [2.25.2]: https://github.com/atlassian/aws-infrastructure/compare/release-2.25.1...release-2.25.2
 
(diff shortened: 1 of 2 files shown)


pczuj:
For some reason when I was reporting the original problem (https://ecosystem.atlassian.net/browse/JPERF-790) I was thinking it fails only for Standalone (I have no idea why).

When I tried to bump the aws-infrastructure to 2.25.2 in our internal tests it started to fail on DC as well.

pczuj:
Again (similar to #123) an automated test for this would be tricky as the problem is seen when logging in to the Jira using web browser. I don't think we should depend on such complex logic inside code.

From my point of view the aws-infrastructure lib should not provide promise that the provisioned Jira will work. It should provide all the possible API points that would allow the lib consumer to make the Jira work. In other words it should provide behavior and not outcome. If that would be the case the problem I'm resolving wouldn't be a bug, but rather missing API.

dagguh:
That's a fair call. We can make it explicit in `DataCenterFormula` Javadoc, or at least in README

dagguh (approved):

pczuj:
Can't merge because the automation claims again that I didn't sign CLA. Not sure what to do.

dagguh:
Options:
* retrigger all checks via Git push
* remove CLA as required check for `master`
* escalate to internal `#opensource` Slack

pczuj:
I will escalate internally

PR #126, opened 2022-01-31
Mark 2.25.3 release

pczuj:


pczuj:
BTW I noticed that Github "releases" don't have many past entries, so I'm not adding those for the releases I triggered as well. If I'd be able to preduct exactly how this "releases" feature works, then maybe I'd fix it.

https://github.com/atlassian/aws-infrastructure/releases

dagguh:
Indeed, I wouldn't add an extra manual step to the release process. We're keeping a CHANGELOG.md, which doesn't have as many features, but it's portable and local.

CHANGELOG.md
```diff
@@ -22,6 +22,9 @@ Dropping a requirement of a major version of a dependency is a new contract.
 
 ## [Unreleased]
```
mzyromski-atlassian, line 23: Drive-by change request: could we make "Unreleased" a link to https://github.com/atlassian/aws-infrastructure/compare/release-2.25.3...master, similarly to other JPT components (e.g. https://github.com/atlassian/ssh-ubuntu/blob/master/CHANGELOG.md)?

mzyromski-atlassian (dismissed):

CHANGELOG.md
pczuj: Seems like it was there before, but was removed accidentally as part of https://github.com/atlassian/aws-infrastructure/commit/e728d568392eca1ea7e5d5ce4ce15e28d7e11170. I will bring it back.

- [ ] Bring back link to Unreleased changes

mzyromski-atlassian (approved):

PR #131, opened 2022-05-11
JPERF-760: Print Jira logs when failed to start a node

pczuj:
Currently if I'm not missing something we are not collecting any artifacts if `InfrastructureFormula` fails to provision (throws an `Exception`). When we fail to wait on Jira node to start then the process throws an `Exception` and in the result makes us unable to diagnose why the failure happened in the first place, because we are not getting handle on `ProvisionedInfrastructure`.

I though that we may want to just print the Jira logs in the JPT's `detailed.log`. It should not cause much harm.

I'm looking for feedback.

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneStoppedNode.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneStoppedNode.kt
@@ -153,6 +153,13 @@ internal data class StandaloneStoppedNode(
                 break
             }
             if (deadline < now()) {
+                if (logger.isDebugEnabled) {
+                    val logContent = ssh.safeExecute(
+                        cmd = "cat $jiraHome/log/atlassian-jira.log",
+                        timeout = ofMinutes(3)
+                    ).output
+                    logger.debug("Jira log file: <<$logContent>>")
+                }
                 throw Exception("$uri failed to get out of $statusQuo status within $timeout")
             }
             threadDump.gather(ssh, "thread-dumps")
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-2.25.5...master
 
+### Added
+- Allow to diagnose the Jira node startup problems by printing atlassian-jira.log inside test log. Resolve [JPERF-760]
+
+[JPERF-760]: https://ecosystem.atlassian.net/browse/JPERF-760
+
 ## [2.25.5] - 2022-04-07
 [2.25.5]: https://github.com/atlassian/aws-infrastructure/compare/release-2.25.4...release-2.25.5
 


src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneStoppedNode.kt
```diff
@@ -153,6 +153,13 @@ internal data class StandaloneStoppedNode(
                 break
             }
             if (deadline < now()) {
+                if (logger.isDebugEnabled) {
+                    val logContent = ssh.safeExecute(
+                        cmd = "cat $jiraHome/log/atlassian-jira.log",
```
mgrzaslewicz, line 158: It would be great to download all rolled logs, not only latest one as the reason of failure might logged somewhere at the beginning

pczuj: Do you have any suggestion how to do that without changing the whole architecture of the lib?

I personally thought that a proper solution would be to not use Exceptions as signals of failure. This would allow us to control artifact gathering even for infrastructure that would fail to provision. 

Yet for such big changes I think we need to do a lot more work. I see it as a next step, unless there is some easier way then the logging or change of Exception APIs.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneStoppedNode.kt
```diff
@@ -153,6 +153,13 @@ internal data class StandaloneStoppedNode(
...
+                        timeout = ofMinutes(3)
+                    ).output
+                    logger.debug("Jira log file: <<$logContent>>")
```
pczuj, line 161: I was thinking about TRACE level of log, however for our internal test usage it doesn't matter and I'm not really sure which one is proper. Do you think DEBUG is fine?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneStoppedNode.kt
mgrzaslewicz: Simple cat command change should be enough
```echo 0 > test.log
echo 1 > test.log.1
echo 2 > test.log.2
cat test.log*
0
1
2

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneStoppedNode.kt
mgrzaslewicz: I think it's fine

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneStoppedNode.kt
pczuj: Is the order of files in such print guaranteed?

I believe Jira's logs roll so that higher number at the end means older log file. If so we would probably want it to be printed the other way around to avoid confusion.

mgrzaslewicz: I'm not sure how it will behave on Jira logs, needs to be tested on the Jira host

pczuj: I see that it would take more of my time to ship this. I will stick to printing only one file for now. Even with printing all of them we won't get full log history, as Jira deletes the oldest files.

We can add support for printing all `atlassian-jira.log*` files later on at any point in the future without breaking anything.

mgrzaslewicz: Having the last log file is still better than having nothing. Agreed, we can improve it later

pczuj:
I was testing this change locally when diagnosing some problem in our internal Jira CI. The log was printed correctly and I was able to deduce where is the start and where is the end of it.

I didn't add any automated tests to changes in this PR as I don't think logging capability is good to be tested while strictly relying on specific logging library. I will skip this.

mgrzaslewicz (approved):

PR #132, opened 2022-05-24
NONE: Update locks

pczuj:
This is mainly to pull in new diagnostics from aws-resources, so that we are able to know why infrastructure failed to provision.


pczuj:
I won't be able to merge it, because CI doesn't work: https://atlassian.slack.com/archives/CFGCT2G49/p1652268362005659

PR #135, opened 2022-07-29
JPERF-821: Introduce integration test that reveals recently changed AMI incompatibility with some regions

pczuj:
This is only a test for now. I'm not sure how to fix it yet.

### DIFF
--- a/src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/AmiNameResolverIT.kt
+++ b/src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/AmiNameResolverIT.kt
@@ -1,0 +1,91 @@
+package com.atlassian.performance.tools.awsinfrastructure
+
+import com.amazonaws.regions.Regions
+import com.amazonaws.regions.Regions.*
+import com.atlassian.performance.tools.aws.api.Aws
+import org.hamcrest.Matchers.isEmptyString
+import org.hamcrest.Matchers.not
+import org.junit.Assert.assertThat
+import org.junit.Test
+
+class AmiNameResolverIT {
+    @Test
+    fun worksInUsEast1() {
+        checkRegion(US_EAST_1)
+    }
+
+    @Test
+    fun worksInUsEast2() {
+        checkRegion(US_EAST_2)
+    }
+
+    @Test
+    fun worksInUsWest1() {
+        checkRegion(US_WEST_1)
+    }
+
+    @Test
+    fun worksInUsWest2() {
+        checkRegion(US_WEST_2)
+    }
+
+    @Test
+    fun worksInEuWest1() {
+        checkRegion(EU_WEST_1)
+    }
+
+    @Test
+    fun worksInEuWest2() {
+        checkRegion(EU_WEST_2)
+    }
+
+    @Test
+    fun worksInEuWest3() {
+        checkRegion(EU_WEST_3)
+    }
+
+    @Test
+    fun worksInEuCentral1() {
+        checkRegion(EU_CENTRAL_1)
+    }
+
+    @Test
+    fun worksInEuNorth1() {
+        checkRegion(EU_NORTH_1)
+    }
+
+    @Test
+    fun worksInApSouth1() {
+        checkRegion(AP_SOUTH_1)
+    }
+
+    @Test
+    fun worksInApSouthEast1() {
+        checkRegion(AP_SOUTHEAST_1)
+    }
+
+    @Test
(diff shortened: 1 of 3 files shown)


src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/AmiNameResolverIT.kt
```diff
@@ -0,0 +1,96 @@
+package com.atlassian.performance.tools.awsinfrastructure
+
+import com.amazonaws.regions.Regions
+import com.amazonaws.regions.Regions.*
+import com.atlassian.performance.tools.aws.api.Aws
+import org.junit.Test
+
+class AmiNameResolverIT {
```
pczuj, line 8: Added a test for every US, EU and AP region. Didn't add for other ones.

On my local `EU_SOUTH_1` (Milan) and `AP_EAST_1` (Hong Kong) show some authentication issue, so we might not want to check on those. I will probably remove those 2 after confirming that they fail similarly on CI.

The only region that really is failing on my local from the ones that I added tests for is `US_EAST_1`, which is very interesting.

pczuj:
Added assert in the checks, so that the test doesn't rely solely on the exceptions not being thrown.

Also removed 2 regions from the new test that I mentioned in https://github.com/atlassian/aws-infrastructure/pull/135#discussion_r933152142. We fail on authentication to those 2 (`EU_SOUTH_1` and `AP_EAST_1`).

pczuj:
Figured out why this is failing and commented in the issue: https://ecosystem.atlassian.net/browse/JPERF-821
> It looks like the problem is because there are 2 images matching the `ubuntu/images/hvm-ssd/ubuntu-focal-20.04-amd64-server-20220610` name and our code uses [`singleOrNull`](https://ecosystem.atlassian.net/browse/JPERF-821) when taking one element out of the returned collection.

src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/AmiNameResolverIT.kt
```diff
@@ -0,0 +1,91 @@
...
+        checkRegion(AP_SOUTHEAST_1)
+    }
+
+    @Test
+    fun worksInApNorthEast2() {
+        checkRegion(AP_SOUTHEAST_2)
+    }
+
+    private fun checkRegion(region: Regions) {
+        val vuAmi = AmiNameResolver.vuAmi(
```
dagguh, line 83: This logic (and test) belongs to [aws-resources](https://bitbucket.org/atlassian/aws-resources).

pczuj:
Forgot to add CHANGELOG entry.

- [ ] Add CHANGELOG.md entry

src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/AmiNameResolverIT.kt
pczuj: I was thinking about it during the weekend and I'd like to skip the correction of this for now. This is mostly because I'm currently not sure how the aws-resources API should look like to properly address aws-infrastructure needs around AMIs.

I will raise an issue to move the AMI search responsibility to the proper place, but for now I need to bring the tests to the operational state without significant delay.

This PR we are commenting in should still improve the state and not make it worse.

dagguh: > I'm currently not sure how the aws-resources API should look like to properly address aws-infrastructure needs around AMIs.

This can stay as is for now, so no "significant delay".

pczuj:
Added the `CHANGELOG` entry. Nothing else changed.

src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/AmiNameResolverIT.kt
dagguh: Ah, I thought that the `AmiNameResolver` is in `aws-resources`, but it's here. Sorry for the confusion.

dagguh (approved):

pczuj:
Yes, `AmiNameResolver` is in `aws-infrastructure`, however it probably should be in `aws-resources`.

In fact the name `AmiNameResolver` is incorrect, because it is resolving AMI's ID and not the name.

Not sure if we shouldn't just hardcode the IDs. Would be more strict and runtime safe. Disadvantage though would be the reduced readability of the code.

Thanks for the review 👍

PR #138, opened 2022-12-15
JPERF-188 Fix stopping Jira while saving a dataset by replacing tomcat shutdown command with 'stop-jira.sh' script.

ewefie:
Using a dedicated script being a part of Jira product is a recommended way to stop an instance.

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/CustomDatasetSource.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/CustomDatasetSource.kt
@@ -1,6 +1,7 @@
 package com.atlassian.performance.tools.awsinfrastructure.api
 
 import com.atlassian.performance.tools.aws.api.StorageLocation
+import com.atlassian.performance.tools.awsinfrastructure.api.jira.Jira
 import com.atlassian.performance.tools.concurrency.api.submitWithLogContext
 import com.atlassian.performance.tools.infrastructure.api.dataset.Dataset
 import com.atlassian.performance.tools.ssh.api.Ssh
@@ -59,14 +60,25 @@ class CustomDatasetSource private constructor(
             .build()
     }
 
+    @Deprecated(message = "Use CustomDatasetSource.storeInS3(location, jira) instead",
+        replaceWith = ReplaceWith("store(location, null)")
+    )
     fun storeInS3(
         location: StorageLocation
     ) {
-        store(location)
+        store(location, null)
+    }
+
+    fun storeInS3(
+        location: StorageLocation,
+        jira: Jira
+    ) {
+        store(location, jira)
     }
 
     internal fun store(
-        location: StorageLocation
+        location: StorageLocation,
+        jira: Jira?
     ): Dataset {
         logger.info("Uploading dataset to '$location'...")
         val executor = Executors.newFixedThreadPool(
@@ -76,7 +88,12 @@ class CustomDatasetSource private constructor(
                 .build()
         )
 
-        stopJira(jiraHome.host)
+        if (jira == null) {
+            @Suppress("DEPRECATION")
+            stopJira(jiraHome.host)
+        } else {
+            stopJira(jira)
+        }
         stopDockerContainers(database.host)
 
         val jiraHomeUpload = executor.submitWithLogContext("jiraHome") {
@@ -109,10 +126,15 @@ class CustomDatasetSource private constructor(
         return DatasetCatalogue().custom(location)
     }
 
+    @Deprecated(message = "Use CustomDatasetSource.stopJira(Jira, JavaDevelopmentKit) instead")
     private fun stopJira(host: SshHost) {
         val shutdownTomcat = "echo SHUTDOWN | nc localhost 8005"
         val waitForNoJavaProcess = "while ps -C java -o pid=; do sleep 5; done"
         Ssh(host, connectivityPatience = 4).newConnection().use { it.safeExecute("$shutdownTomcat && $waitForNoJavaProcess", Duration.ofMinutes(3)) }
+    }
+
+    private fun stopJira(jira: Jira) {
+        jira.stop()
     }
 
     private fun stopDockerContainers(host: SshHost) {
--- a/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/StartedNode.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/StartedNode.kt
@@ -3,6 +3,7 @@ package com.atlassian.performance.tools.awsinfrastructure.api.jira
 import com.atlassian.performance.tools.aws.api.Storage
 import com.atlassian.performance.tools.awsinfrastructure.api.aws.AwsCli
 import com.atlassian.performance.tools.infrastructure.api.jira.JiraGcLog
+import com.atlassian.performance.tools.infrastructure.api.jvm.JavaDevelopmentKit
 import com.atlassian.performance.tools.infrastructure.api.process.RemoteMonitoringProcess
 import com.atlassian.performance.tools.ssh.api.Ssh
 import java.time.Duration
@@ -14,7 +15,8 @@ class StartedNode(
     private val resultsTransport: Storage,
     private val unpackedProduct: String,
     private val monitoringProcesses: List<RemoteMonitoringProcess>,
-    private val ssh: Ssh
+    private val ssh: Ssh,
+    private val jdk: JavaDevelopmentKit
 ) {
     private val resultsDirectory = "results"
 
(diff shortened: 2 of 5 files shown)


src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/CustomDatasetSource.kt
```diff
@@ -59,14 +60,25 @@ class CustomDatasetSource private constructor(
             .build()
     }
 
+    @Deprecated(message = "Use CustomDatasetSource.storeInS3(location, jira) instead",
+        replaceWith = ReplaceWith("store(location, null)")
```
pczuj, line 64: If `store(location, null)` is `internal` would that work?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/StartedNode.kt
```diff
@@ -68,9 +70,20 @@ class StartedNode(
             resultsTransport = this.resultsTransport,
             unpackedProduct = this.unpackedProduct,
             monitoringProcesses = this.monitoringProcesses,
-            ssh = this.ssh
+            ssh = this.ssh,
+            jdk = this.jdk
         )
     }
 
+    fun stopNode() {
```
pczuj, line 78: I though: Should it return `StoppedNode`? That would kinda introduce cyclic dependency, however would follow the architecture of returning up to date client related to the state after operation.

In the end I don't see it as a big deal, however I thought it's worth to highlight this puzzle.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/StartedNode.kt
```diff
@@ -68,9 +70,20 @@ class StartedNode(
...
+        ssh.newConnection()
+            .use { ssh ->
+                ssh.safeExecute(
+                    "${jdk.use()}; ./${unpackedProduct}/bin/stop-jira.sh",
```
dagguh, line 82: Oh snap, `jdk` is needed for this? I thought it's all bash scripts.
Unfortunately, adding this to the constructor breaks old constructor calls. Sure, from `ijpt` perspective we only see it called from awsinfra, so in theory it wouldn't break any caller. But it's a nasty surprise for anybody in another repo or another branch. And precedent for a dangerous shortcut.

First, I'd see if it can just skip the `jdk.
If not, then let's hack the impl instead of hacking the API:
1. keep the new primary constructor
2. restore the old constructor as a secondary constructor, deprecate it
3. use a "default JDK" in the old constructor, even lazy-install it if it's necessary

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/StartedNode.kt
dagguh: That would be nice to have. But the `StoppedNode` would need to be passed via the constructor. Every time we expose a constructor it's a huge pain. We probably need builders every time. AWS SDK does exactly that.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/CustomDatasetSource.kt
```diff
@@ -59,14 +60,25 @@ class CustomDatasetSource private constructor(
...
     fun storeInS3(
         location: StorageLocation
     ) {
-        store(location)
+        store(location, null)
+    }
+
+    fun storeInS3(
+        location: StorageLocation,
+        jira: Jira
```
pczuj, line 74: I think jira should be passed in the constructor and not here. The operation so far accepted "where to store", which was enough as the interface for "DatasetSource". `jira` represents "what to do before store" and I don't think that will change between different `storeInS3` executions.

Is there some technical reason why we can't pass `jira` in the constructor (`Builder`)? 

Maybe we could even skip the `if (jira == null)` if we'd model the dependency as "pre store operation" or just something like "jira stopper", "jira stop handler", "jira hand break"

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/StartedNode.kt
dagguh: Or maybe Java 9 module system could hide that stuff.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/StartedNode.kt
pczuj, line 82: QA check: Will that succeed if called twice? i.e. when we store dataset twice.

ewefie: I already tried without `jdk` and got this message: "No JRE_HOME or JAVA_HOME environment variable is set - attempting to just run java command" but I need to verify if it failed or maybe somehow succeeded.

dagguh: It's `safeExecute`, so it will work even if `stop-jira.sh` is not idempotent

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/CustomDatasetSource.kt
dagguh: Good idea. Pass it via builder.
But the `jirahome` version is a bug, doesn't work for DC, better delete it altogether. Only the new calls with `Jira` will work properly. We can signal that via API:
1. make `Jira` a required parameter in the `Builder`
2. deprecate the old builder constructor in favor of the new one

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/StartedNode.kt
pczuj: We can do optional `jdk`, i.e. nullable internally, and throw / return "no success" when we don't have it. It's a new feature that is supported only if you use the new builder constructor.

dagguh (changes requested):
I'm frustrated that those APIs are so hard to change. I apologize for leaving such unmaintainable code.
First, let's try the compatibility-preserving solutions we commented about.
If that costs too many complications, then let's just break it and do the major release.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/CustomDatasetSource.kt
dagguh: Indeed it wouldn't

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/CustomDatasetSource.kt
```diff
@@ -109,12 +126,17 @@ class CustomDatasetSource private constructor(
         return DatasetCatalogue().custom(location)
     }
 
+    @Deprecated(message = "Use CustomDatasetSource.stopJira(Jira, JavaDevelopmentKit) instead")
```
pczuj, line 129: Do we need to deprecate `private` methods? Without it we can also remove the `Suppress` from current line 92

dagguh: `@Deprecate` means don't use this
So if a `private` thing should not be used, then it's a proper annotation (doesn't affect API tho, so nothing in the changelog).
Naturally, `private` things are easy to change, so by default we remove it and use the replacement. But in general, there could be situations where making the replacement is hard right now, so I guess those could stay.

pczuj: Makes sense

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/StartedNode.kt
ewefie: Ok, I have checked how it works without JDK and even though no explicit error is thrown, the script doesn't work. After execution, Jira is still running.

dagguh: Noice :muscle: 
Also, we're supposed to simulate a Jira admin. I doubt they write `JAVA_HOME=dfhsdkfhsdkf ./bin/start-jira.sh` every time. I bet they configure it once in `~/.bashrc` during JDK/Jira installation.

pczuj: Yup, we probably should register the env variable for every process as part of Java installation

ewefie:
Closed in favor of: https://github.com/atlassian/aws-infrastructure/pull/162

PR #140, opened 2023-01-24
JPERF-831 Migrate CI from CircleCI to Github Actions.

ewefie:
This PR is mostly about moving the CI from CircleCi to GithubActions, however, to confirm everything works correctly I had to bump also `infrastructure` and `aws-infrastructure` libs.
Changes in those libs fix problems with instance provisioning caused by the old Ubuntu image (Xenial). Now instances run on Focal.

cc/ @Nubzor


.github/workflows/build.yml
```diff
@@ -0,0 +1,81 @@
+name: build
+on:
```
ewefie, line 2: I'm not sure what kind of triggers we want to keep. Especially I'm not convinced if we need `on push` on. WDYT?

.github/workflows/build.yml
```diff
@@ -0,0 +1,81 @@
...
+      fail-fast: false
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Configure AWS credentials
```
ewefie, line 26: For AWS auth I used the official [AWS action](https://github.com/aws-actions/configure-aws-credentials). With this approach, we don't have to configure additional identity providers in the code.

.github/workflows/build.yml
```diff
@@ -0,0 +1,81 @@
...
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
+          aws-region: eu-west-1
+      - name: Compile
```
ewefie, line 31: I used [the official Gradle action](https://github.com/gradle/gradle-build-action) here and the reason for that is that it handles caches for us so the whole pipeline configuration is cleaner and shorter (no need for a separate `cache gradle` step as it was in the CircleCi config).

.github/workflows/build.yml
```diff
@@ -0,0 +1,81 @@
...
+        if: always()
+        with:
+          name: test-results
+          path: ./build/test-results
+  build-check:
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - run: echo "All build jobs successful."
+  release:
```
ewefie, line 58: The release part is configured in the same manner as it is done i.e. in [jira-actions](https://github.com/atlassian/jira-actions/blob/master/.github/workflows/ci.yml#L46) lib.

CONTRIBUTING.md
```diff
@@ -58,7 +58,7 @@ those contributing as an individual.
 
 ### Pull Request Checks
 
-Pull requests must pass a [CircleCI](https://circleci.com) build. This is **not** automatically run under Atlassian's account, as contributor you will need to setup and run the CI build against your own fork.
+Pull requests must pass a [Github Actions](https://github.com/atlassian/aws-infrastructure/actions) build. This is **not** automatically run under Atlassian's account, as contributor you will need to setup and run the CI build against your own fork.
```
ewefie, line 61: This part might require updating depending on what we decide regarding triggers.

.github/workflows/build.yml
```diff
@@ -0,0 +1,81 @@
...
+    inputs:
+      release:
+        description: 'Release? yes/no'
+        default: 'no'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
```
pczuj, line 17: Why do we want/need this? 

Why the question: In `jira-actions` we are using it to allow other parallel jobs to finish, so that we have results from other CI-level parametrised tests. Here I don't see us using `strategy.matrix`.

BTW I personally think we shouldn't setup parallelism on the CI level. It leads to differentiation of CI from local devloop which has it's negative consequences. Possibly we'd be able to shift it in `jira-actions` back to Gradle/junit level.

pczuj (approved):

ewefie:
I didn't add a new entry to the changelog as there are no changes in the production code besides a few lib bumps, but I'm okay with adding something if you feel it'd bring some value.

.github/workflows/build.yml
ewefie: Oh, this change was not intentional. I'll remove it.

pczuj (approved):

.github/workflows/build.yml
dagguh: For us `on push` only runs after PR merge. Indeed it's kind of already tested by the `pull_request` trigger (during PR).
*But*, the post-merge code can differ from PR code, because GitHub can rebase for us. I'd leave it in.

.github/workflows/build.yml
dagguh: That's nice, less boilerplate FTW

dagguh (approved):

PR #141, opened 2023-01-27
JPERF-945: Reproduce the Kotlin metadata bug

dagguh:
Rewrite locks: `./gradlew dep --write-locks`.
As you can see, the newest `aws-resources` in the declared range _is_ `1.10.1`. The "is" means "as of the time of this commit",
so you might have a different result if you repeat it later on. Luckily dependency locks let us freeze this moment in time.

To see the broken compatibility, run `./gradlew compileTestKotlin`.


dagguh:
Oops, credentials are now checked before unit tests, so it doesn't show the symptom I need

dagguh:
[Reopened in upstream](https://github.com/atlassian/aws-infrastructure/pull/142) to piggyback on the AWS credentials

PR #142, opened 2023-01-27
JPERF-945: Reproduce the Kotlin metadata bug

dagguh:
Rewrite locks: `./gradlew dep --write-locks`.
As you can see, the newest `aws-resources` in the declared range _is_ `1.10.1`. The "is" means "as of the time of this commit",
so you might have a different result if you repeat it later on. Luckily dependency locks let us freeze this moment in time.

To see the broken compatibility, run `./gradlew compileTestKotlin`.

### DIFF
--- a/gradle/dependency-locks/apiDependenciesMetadata.lockfile
+++ b/gradle/dependency-locks/apiDependenciesMetadata.lockfile
@@ -13,7 +13,7 @@ com.amazonaws:aws-java-sdk-sts:1.11.817
 com.amazonaws:aws-java-sdk-support:1.11.817
 com.amazonaws:jmespath-java:1.11.817
 com.atlassian.data:random-data:1.4.3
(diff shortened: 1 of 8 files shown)


dagguh:
[CI](https://github.com/atlassian/aws-infrastructure/actions/runs/4026602083/jobs/6921311093) confirms the symptom:
```
e: Incompatible classes were found in dependencies. Remove them from the classpath or use '-Xskip-metadata-version-check' to suppress errors
e: /home/runner/.gradle/caches/modules-2/files-2.1/com.atlassian.performance.tools/aws-resources/1.10.1/8ee7bbbc5902d014215c4fbf0b6189f993dfbcc4/aws-resources-1.10.1.jar!/META-INF/aws-resources.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.4.3, expected version is 1.1.11.
```

PR #147, opened 2023-02-02
Random change for testing purpose

ewefie:
I'm opening this PR just to see if the tests will pass.


ewefie:
The pipeline has failed. The error message says "Credentials could not be loaded, please check your action inputs: Could not load credentials from any providers". That's what we expected.

PR #148, opened 2023-02-06
JPERF-968 Add ability to use a long term AWS credentials in the pipelines

ewefie:
Replace the hardcoded region with environment variable.

The preferred way for authenticating in AWS is using IAM role, but 3rd party contributors might prefer using a key pair and with the previous config the key pair would not be picked up in the pipeline.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,9 @@ jobs:
         uses: aws-actions/configure-aws-credentials@v1
         with:
           role-to-assume: ${{ secrets.AWS_ROLE_TO_ASSUME }}
-          aws-region: eu-west-1
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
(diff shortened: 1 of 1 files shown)


ewefie:
Currently the build fails, but the failure is not related to the changes from this PR as it occurs also on master branch and locally. I'm still investigating it.

ewefie:
> Currently the build fails, but the failure is not related to the changes from this PR as it occurs also on master branch and locally. I'm still investigating it.

Well, I don't know what was the reason of that failures. I got fails on master both on CI and locally, and now they are gone 🤔 🤷🏻‍♀️

dagguh:
We had apt-get flakiness yesterday. It was also timing out, making CI sometimes green sometimes red: https://github.com/atlassian/infrastructure/pull/131#issuecomment-1419366653

dagguh (approved):
Good change!

PR #149, opened 2023-02-09
JPERF-997: network stack housekeeping

dagguh:
The main reason for this PR is `NetworkFormulaIT` as a way to test network creation and housekeeping without spinning up unnecessary EC2 instances.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,9 @@ Dropping a requirement of a major version of a dependency is a new contract.
 
 ### Added
 - Return the AWS `Resource` allocated by `NetworkFormula` as `ProvisionedNetwork`.
-- Open `Network` for inheritance.
+
+### Deprecated
+- Deprecate `NetworkFormula.provision` in favor of `NetworkFormula.provisionAsResource`.
 
 ### Fixed
 - Substitute the lost `DatasetCatalogue().largeJiraEight()` with an equivalent. Fix [JPERF-980].
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-2.26.0...master
 
+### Added
+- Return the AWS `Resource` allocated by `NetworkFormula` as `ProvisionedNetwork`.
+- Open `Network` for inheritance.
+
 ### Fixed
 - Substitute the lost `DatasetCatalogue().largeJiraEight()` with an equivalent. Fix [JPERF-980].
   JQL data might be bigger than in the original. Requires Jira v8.22.0 or higher.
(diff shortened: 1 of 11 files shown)


dagguh:
I already used this test to [confirm that "network stack" housekeeping works properly](https://ecosystem.atlassian.net/browse/JPERF-997?focusedCommentId=925921). I don't need it merged, but wanted to share the dev loop piece. PS. it also fixes some minor (but annoying) pain points.

dagguh:
Rebased onto master (after #151) to get the CI fix.

dagguh:
Now with a rebase onto the proper fix from #152

dagguh:
Rebased onto master (#152)

dagguh:
@atlassian/jpt review pls

ewefie (approved):

CHANGELOG.md
```diff
@@ -23,9 +23,18 @@ Dropping a requirement of a major version of a dependency is a new contract.
...
+- Return the AWS `Resource` allocated by `NetworkFormula` as `ProvisionedNetwork`.
+
+### Deprecated
+- Deprecate `NetworkFormula.provision` in favor of `NetworkFormula.provisionAsResource`.
+
 ### Fixed
 - Substitute the lost `DatasetCatalogue().largeJiraEight()` with an equivalent. Fix [JPERF-980].
   JQL data might be bigger than in the original. Requires Jira v8.22.0 or higher.
+- Track AWS resources from `ProvisionedNetwork` for release.
+- Stop spamming INFO log with network internals.
```
pczuj, line 36: Nice!

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/network/NetworkFormula.kt
```diff
@@ -34,7 +35,22 @@ class NetworkFormula(
...
             stack.findSubnet("TheOnlySubnet")
         )
-        logger.info("Network provisioned: $network")
-        return network
+        logger.info("Network provisioned")
+        logger.debug("Network provisioned: $network")
+        return ProvisionedNetwork(network, stack)
     }
+
+    internal fun reuseOrProvision(existing: Network?) : ProvisionedNetwork {
```
pczuj, line 43: I'll be honest that I don't like the idea that this method exists. Good that it's `internal`, so that we can refactor it later.

I don't want to go deeper why existance of this method is a problem, because there is a design level flaw one step higher. I think we should at some point get rid of the hacky "network override" mechanism done inside `InfrastructureFormula`. It relies on `instanceof`, which makes it tricky to use custom implementations of `JiraFormula`.

Hopefully API redesign as part of https://ecosystem.atlassian.net/browse/JPERF-273 will cover this problem.

dagguh: True, it's smelly. The network feature added this complication due to compatibility. It should be simplified in the next major release. It's now easier to track this smell, because it's extracted and named.

PR #151, opened 2023-02-10
JPERF-980: Substitute the lost Jira 8 dataset

dagguh:


pczuj (approved):

dagguh:
This was red and got merged? What the...

dagguh:
I overrode the branch protection to drop the bad red commit from master. Proper fix in #152. Let's see if the builds will be considered as `Required`.

PR #152, opened 2023-02-10
JPERF-980: Substitute the lost Jira 8 dataset

dagguh:
Unfortunately, the new dataset doesn't work for Jira v8.0.0, it needs v8.22.0 or higher.
The lost dataset used to support v8.0.0.

It wasn't an explicit contract, so technically this isn't breaking declared compatibility. However, it was relied upon, so it was a de facto implicit contract.
So we hope to restore v8.0.0, and add that as a new explicit contract.


dagguh:
Hmm, the `build-check` is `Required`, so that looks good.
image

ewefie (approved):

PR #154, opened 2023-02-22
JPERF-996: Make it possible to handle waiting for Jira startup outside provisioning process

pczuj:
When Jira is set up from scratch we can't rely on the current version of wait mechanism as it assumes Jira will run upgrade tasks before it's interactive.

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
@@ -56,7 +56,8 @@ class DataCenterFormula private constructor(
     private val databaseComputer: Computer,
     private val databaseVolume: Volume,
     private val accessRequester: AccessRequester,
-    private val adminPasswordPlainText: String
+    private val adminPasswordPlainText: String,
+    private val waitForUpgrades: Boolean
 ) : JiraFormula {
     private val logger: Logger = LogManager.getLogger(this::class.java)
 
@@ -88,7 +89,8 @@ class DataCenterFormula private constructor(
         databaseComputer = M4ExtraLargeElastic(),
         databaseVolume = Volume(100),
         accessRequester = Defaults.accessRequester,
-        adminPasswordPlainText = Defaults.adminPasswordPlainText
+        adminPasswordPlainText = Defaults.adminPasswordPlainText,
+        waitForUpgrades = true
     )
 
     @Suppress("DEPRECATION")
@@ -111,7 +113,8 @@ class DataCenterFormula private constructor(
         databaseComputer = M4ExtraLargeElastic(),
         databaseVolume = Volume(100),
         accessRequester = Defaults.accessRequester,
-        adminPasswordPlainText = Defaults.adminPasswordPlainText
+        adminPasswordPlainText = Defaults.adminPasswordPlainText,
+        waitForUpgrades = true
     )
 
     override fun provision(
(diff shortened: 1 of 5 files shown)


pczuj:
I used this in empty data generation. Modeled the mechanism using "EmptyJiraFormula" implementing `JiraFormula` which I based on `StandaloneFormula`, however with self-initiating MySQL database and empty jirahome directory.

This change is needed for any Jira that strats in setup mode, i.e. you can't log into it yet.

Looking for ideas how to implement it better.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneStoppedNode.kt
```diff
@@ -107,6 +108,11 @@ internal data class StandaloneStoppedNode(
         ssh: SshConnection,
         threadDump: ThreadDump
     ) {
+        if (skipStartupWait) {
+            logger.info("Skipping Jira startup wait")
+            return
+        }
+        logger.info("Waiting for Jira node to start...")
```
pczuj, line 115: Logging might be too verbose? Would `debug` be more appropriate or `info` is fine?

I pushed just what I had for now. Am willing to change when decided.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
```diff
@@ -420,7 +425,9 @@ class DataCenterFormula private constructor(
             network = formula.overriddenNetwork
             databaseComputer = formula.databaseComputer
             databaseVolume = formula.databaseVolume
+            accessRequester = formula.accessRequester
             adminPasswordPlainText = formula.adminPasswordPlainText
+            skipStartupWait = formula.skipStartupWait
```
pczuj, line 430: This part is easy to miss. The code seem work without it, but will break when used with `InfrastructureFormula` and the option will be overwritten back to `false`.

dagguh: Yes, it kinda easy to forget about copying constructors

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneStoppedNode.kt
```diff
@@ -107,6 +108,11 @@ internal data class StandaloneStoppedNode(
         ssh: SshConnection,
         threadDump: ThreadDump
     ) {
+        if (skipStartupWait) {
```
dagguh, line 111: We can reuse the existing terminology:
```suggestion
        if (!waitForUpgrades) {
```

dagguh (dismissed):

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneStoppedNode.kt
pczuj: Applied this suggestion

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/StandaloneStoppedNode.kt
pczuj: Applied this suggestion

dagguh (approved):

PR #156, opened 2023-03-15
JPERF-990: Make it possible to override `AwsDatasetModification`'s timeouts of storing datasets

pczuj:
Didn't want to duplicate the timeouts definitions in multiple places.


pczuj:
Not sure if this change makes sense. Please suggest any other solution for allowing the consumer to define the timeouts.

pczuj:
Things I can think about:
 * Maybe we don't want to have this configurable and we want to hardcode long timeout in the implementation?
 * Maybe we want to expose the `CustomDatasetSource.Builder` instead?
 * Maybe we want to expose every timeout from `CustomDatasetSource` value as separate parameter of `AwsDatasetModification`? What about the lower level defaults in such case? Are we replacing them with overwritten ones?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
```diff
@@ -132,6 +132,13 @@ class AwsDatasetModification private constructor(
                 )
                 .build()
         }
+        private var customDatasetSourceFactory: CustomDatasetSourceFactory = object : CustomDatasetSourceFactory {
+            override fun createCustomDatasetSource(
+                jiraHome: RemoteLocation,
+                database: RemoteLocation
+            ) = CustomDatasetSource.Builder(jiraHome, database)
```
dagguh, line 139: I was thinking about a `Consumer<CustomDatasetSource.Builder>` instead of a factory. It would make wrapping `jiraHome` and `database` harder, but it's already possible via `JiraFormula` via `DatasetHost`.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/CustomDatasetSourceFactory.kt
```diff
@@ -0,0 +1,8 @@
+package com.atlassian.performance.tools.awsinfrastructure.api
+
+interface CustomDatasetSourceFactory {
```
dagguh, line 3: If the factory has to exist, then let's move it next into `CustomDatasetSource`, just like its `Builder`.

CHANGELOG.md
```diff
@@ -23,6 +23,12 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-2.28.0...master
 
+### Added
+- Make it possible to override `AwsDatasetModification`'s timeouts of storing datasets. Unblock [JPERF-990].
+- Intoduce `CustomDatasetSourceFactory` as an SPI of `AwsDatasetModification`.
```
dagguh, line 28:
```suggestion
- Introduce `CustomDatasetSourceFactory` as an SPI of `AwsDatasetModification`.
```

CHANGELOG.md
```diff
@@ -23,6 +23,12 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-2.28.0...master
 
+### Added
+- Make it possible to override `AwsDatasetModification`'s timeouts of storing datasets. Unblock [JPERF-990].
```
dagguh, line 27: Good :+1:

dagguh (changes requested):

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
pczuj: Sounds like a multi-layer builder pattern that I was exploring in our internal analytics module. I will be honest that I'm not the biggest fan of using builders in contexts other than as a replacement for ton of different constructors, however this pattern may be unavoidable in a certain scale of software. 

Anyway I trust your judgement as an JPT architect.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/CustomDatasetSourceFactory.kt
pczuj: I will remove the factory as you suggested in the other comment, however just hypothetically I'm not sure how this move should be implemented. Would the factory still be an interface? Would it be `CustomDatasetSource.Factory`? I guess the interface method could evolve together with the constructor of `CustomDatasetSource.Builder`.

dagguh: > Would it be `CustomDatasetSource.Factory`?

Yes. Same type, different location.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/AwsDatasetModification.kt
dagguh: Factories are also _mainly_ to replace constructors. But as we can see here, both factories and builders are useful in other circumstances too.

dagguh: `datasetSourceConfig` sounds good

dagguh (dismissed):

CHANGELOG.md
```diff
@@ -23,6 +23,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-2.28.0...master
 
+### Added
```
dagguh, line 26: Oh snap. My [random PR](https://github.com/atlassian/aws-infrastructure/pull/137) caused a conflict here, sorry

dagguh (approved):

PR #161, opened 2023-04-06
Change default virtual user instance type to `c5.9xlarge`

pczuj:
It's better, cheaper and there seem to be availability issues with previous default (`c4.8xlarge`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-2.29.0...master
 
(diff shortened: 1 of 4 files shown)


pczuj:
Does it make sense? Is there any risk of this change?

pczuj:
I thought we made this change before, but we only [applied it in our internal tests](https://stash.atlassian.com/projects/JIRASERVER/repos/jira-performance-tests/pull-requests/1073/overview). Making it a default will be easier to apply it wherever old default wasn't overwritten.

dagguh (approved):
Good stuff. Changing instance types can cause EC2 limit problems but:
* c5.9xlarge is already widely used, so users already have bumped limits
* `CapacityMediator` is available to bump those automatically

PR #162, opened 2023-04-27
JPERF-1079 Make it possible to stop Jira nodes with a dedicated script.

ewefie:
- Remove deprecated constructors of Jira and CustomDatasetSource. 
- Add to Jira information about its nodes.
- Add the ability to stop Jira nodes using a dedicated script (this part requires having JAVA_HOME set)

PR #163, opened 2023-05-11
JPERF-1106: Gather Apache2 logs

pczuj:


pczuj:
This requires at least 4 changes:
 * Get rid of problematic API changes, e.g. `Jira` 2 new constructors are added. We want to wait for https://github.com/atlassian/aws-infrastructure/pull/162, as it introduces `Builder` for `Jira`
 * Need a test
 * Need to find better names for new classes
 * Need a CHANGELOG entry

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
```diff
@@ -382,16 +384,18 @@ class DataCenterFormula private constructor(
...
-                sharedHome.get().remoteSharedHome
-            ),
-            database = databaseDataLocation,
-            address = loadBalancer.uri,
-            jmxClients = jiraNodes.mapIndexed { i, node -> configs[i].remoteJmx.getClient(node.publicIpAddress) }
+        val jiraHomeLocation = RemoteLocation(
+            sharedHomeSsh.host,
+            sharedHome.get().remoteSharedHome
         )
+        val loadBalancerResultsSource = (provisionedLoadBalancer.loadBalancer as? DiagnosableLoadBalancer)
```
pczuj, line 391: First time used this `as?` syntax. Casts are usually bad, however given the circumstances of our current model I think this is a better solution than putting a new field into `ProvisionedLoadBalancer` and I'm also not certain if adding methods to `LoadBalancer` would be a good idea. Extending the interface and detecting it allows for wrapping of `DiagnosableLoadBalancer` implementations without losing the feature.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/Jira.kt
```diff
@@ -1,41 +1,113 @@
...
+            measurementSources.map { executor.submitWithLogContext("gather $it") { it.gatherResults() } }
+                .map { Runnable { it.get() } }
+        ).run()
+
         executor.shutdownNow()
     }
 
     override fun toString() = "Jira(address=$address)"
+
+    class Builder(
```
pczuj, line 82: I thought that I should probably use the same required constructor params as in https://github.com/atlassian/aws-infrastructure/pull/162/files#diff-100d5a9943e653f7807abf68bf52b9fb2d04cb317126a4b1b4fec5f3d014f03d, however I really wanted to make optional as many of them as possible, so that the consumer could build `Jira` easier, e.g. see https://github.com/atlassian/aws-infrastructure/pull/163/files#diff-bd4b59a5570c8b384c9626a7e8ea1f86f6d1cab564e86aa1f77667a383b731ff

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/SuppressingRunnable.kt
```diff
@@ -0,0 +1,26 @@
+package com.atlassian.performance.tools.awsinfrastructure
+
+internal class SuppressingRunnable(
```
pczuj, line 3: Name suggestions welcome

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/DiagnosableLoadBalancer.kt
```diff
@@ -0,0 +1,8 @@
+package com.atlassian.performance.tools.awsinfrastructure.api.loadbalancer
+
+import com.atlassian.performance.tools.aws.api.StorageLocation
+import com.atlassian.performance.tools.infrastructure.api.loadbalancer.LoadBalancer
+
+interface DiagnosableLoadBalancer : LoadBalancer {
```
pczuj, line 6: Name suggestions welcome

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/DiagnosableLoadBalancer.kt
```diff
@@ -0,0 +1,8 @@
...
+    fun gatherDiagnostics(location: StorageLocation)
```
pczuj, line 7: Name suggestions welcome

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/loadbalancer/DiagnosableLoadBalancer.kt
pczuj, line 6: We could make that `LoadBalancer`-agnostic, however that would make the model more abstract and that possibly could make it harder to understand.

src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/SuppressingRunnableTest.kt
```diff
@@ -0,0 +1,69 @@
...
+        val runnable = SuppressingRunnable(
+            listOf(
+                Runnable { throw Exception("Banana") },
+                Runnable { throw Exception("Apple") },
+                Runnable { throw Exception("Pear") },
+                Runnable { throw Exception("Peach") }
+            )
+        )
+
+        val exception = try {
```
ewefie, line 43: Nit:
```
        val exception = catchThrowable { runnable.run() }

        assertThat(exception)
            .isInstanceOf(Exception::class.java)
            .hasMessage("Banana")

        assertThat(exception?.suppressed)
            .extracting("message")
            .containsExactly("Apple", "Pear", "Peach")`

File: src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/SuppressingRunnableTest.kt [hunk 7]
```diff
@@ -0,0 +1,69 @@
...
+        var executed3 = false
+        val runnable = SuppressingRunnable(
+            listOf(
+                Runnable { executed1 = true; throw Exception("Fail 1") },
+                Runnable { executed2 = true; throw Exception("Fail 2") },
+                Runnable { executed3 = true; throw Exception("Fail 3") }
+            )
+        )
+
+        try {
```
ewefie, line 21: Nit (would require bumping JUnit to v4.13):
`assertThrows(Exception::class.java) { runnable.run() }
`

src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/SuppressingRunnableTest.kt
pczuj: I wanted to avoid explicit assert on what message lands in suppressed and what in the "root", because this is not important - the important fact is that all of the messages (and actually also stacktraces, but I don't know how to assert on that reliably) are available inside the single exception in some way.

I will see how to use AssertJ for that.

src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/SuppressingRunnableTest.kt
pczuj: Not sure if I want to assert on the exception here. I believe such assert would be better to be in another test. That's esspecially because adding another unit test has negligible execution cost and is able to separate requirements.

pczuj: If we want to reduce this line to single one we could also use Kotlin's [runCatching](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/run-catching.html), however it would also require dependency bump, as it's available from Kotlin 1.3 and we have 1.2 (IIRC there is some compatibility reason why we are not up to date).

I'm not sure if it's worth the effort.

pczuj:
I migrated the new tests to use AssertJ

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/SuppressingRunnable.kt
pczuj: Suggestin from pairing with @ewefie: `FailSafeRunnable`

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/Jira.kt
```diff
@@ -1,41 +1,113 @@
...
+ * or anything that can be integrated with Jira as part of web application provisioning
+ */
+class Jira private constructor(
     private val nodes: List<StartedNode>,
     val jiraHome: RemoteLocation,
     val database: RemoteLocation?,
     val address: URI,
-    val jmxClients: List<JmxClient> = emptyList()
+    val jmxClients: List<JmxClient>,
+    private val extraResultsSources: List<MeasurementSource>
```
pczuj, line 24: Sggestion from pairing with @ewefie: `extraMeasurementSource`

and propagate to all new API

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/Jira.kt
pczuj: We paired on this with @ewefie and we decided that we don't want to merge it without @dagguh's review.

@dagguh this is the main uncertainity we have in this PR. Please share your point of view.

pczuj:
This is ready for review again. I'm not planning to add integration test for the logs.

CHANGELOG.md
```diff
@@ -23,6 +23,19 @@ Dropping a requirement of a major version of a dependency is a new contract.
...
 
+## Added
+- Add `DiagnosableLoadBalancer`.
+- Add ability to `gatherEvidence` from `ApacheProxyLoadBalancer` by making it `DiagnosableLoadBalancer`.
+- Make it possible to specify `extraMeasurementSources` inside `Jira`, so that it's possible to `gatherResults` from any of Jira hosted integrations, e.g. load balancer or plugins.
+- Let `Jira` produced by `DataCenterFormula` gather logs of `ApacheProxyLoadBalancer`. Resolve [JPERF-1106].
+- Make `StartedNode` a `MeasurementSource`.
+
+## Fixed
+- Let every `MeasurementSource` inside `Jira` finish its gathering even if one of them fails. Fix [JPERF-1114].
```
dagguh, line 34: Let's split those changes. One sounds good, while the other needs some further considerations.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt
dagguh: Indeed casts are bad. See https://github.com/atlassian/aws-infrastructure/blob/release-2.29.0/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormula.kt#L360:
```kotlin
if (loadBalancer is ApacheProxyLoadBalancer) listOf(loadBalancer::updateJiraConfiguration) else emptyList()
```
Fortunately, `ApacheProxyLoadBalancer` will match both `is`/`as?` casts, so it will function.
We need hooks API to avoid further tight coupling. If the apache logs are necessary, then we can merge the tech debt. But if not, then let's skip that part.

dagguh (changes requested):

pczuj:
> If the apache logs are necessary, then we can merge the tech debt. But if not, then let's skip that part.

Not sure how to judge that.

We wanted to take a look at the appache logs when diagnosing a problem recently. We used JAR built from this code to do so, however didn't really find anything useful other than confirmation that the load balancer just times out on communication with the product.

We no longer have an immediate problem to diagnose, so that would mean that the logs are not necessary. Question is if we wouldn't waste the effort if the diagnostic need will repeat itself.

PR #164, opened 2023-05-19
Let `StartedNode.gatherResults` finish despite failure in `stop` of `RemoteMonitoringProcess`

pczuj:
I know it would be better to just fix the `stop` for every `RemoteMonitoringProcess`, however I don't think it should prevent all results gathering if only the `RemoteMonitoringProcess` is faulty. Especially that [we don't fail a test if the results are not gathered](https://github.com/atlassian/aws-infrastructure/blob/c10ced8d753892203c3630bafeb3f1eec491ff89/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/Infrastructure.kt#L68).

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/StartedNode.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/StartedNode.kt
@@ -5,6 +5,7 @@ import com.atlassian.performance.tools.awsinfrastructure.api.aws.AwsCli
 import com.atlassian.performance.tools.infrastructure.api.jira.JiraGcLog
 import com.atlassian.performance.tools.infrastructure.api.process.RemoteMonitoringProcess
 import com.atlassian.performance.tools.ssh.api.Ssh
+import org.apache.logging.log4j.LogManager
 import java.time.Duration
 
 class StartedNode(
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/StartedNode.kt
```diff
@@ -16,11 +17,20 @@ class StartedNode(
...
     private val resultsDirectory = "results"
 
     fun gatherResults() {
         ssh.newConnection().use { shell ->
-            monitoringProcesses.forEach { it.stop(shell) }
+            val successfullyStoppedProcesses = monitoringProcesses.mapNotNull {
+                try {
+                    it.stop(shell)
+                    it
+                } catch (e: Exception) {
```
pczuj, line 29: Maybe we should also throw all exceptions at the end of `gatherResults`?

dagguh:
If a specific `RemoteMonitoringProcess` is flaky, then let's fix that one.

pczuj:
We talked about this on VC and @dagguh noticed that it's possible to just disable `RemoteMonitoringProcess`s in a test whenever it's unstable.

@ewefie did that in out internal experiment.

I'm declining this PR.

PR #166, opened 2023-05-24
Allow storing Avatars/Attachments in S3 

MatthewCochrane:
Adds a parameter to DataCenterFormula to store avatar and/or attachment data in an S3 bucket.  For use with the Object Storage for Jira projects for [avatars](https://team.atlassian.com/project/ATLAS-32366/about) and [attachments](https://team.atlassian.com/project/ATLAS-36873/updates).

It uses a new cloud formation stack for the bucket which gets set up with appropriate permissions (based on the account, which is why I created [this PR](https://github.com/atlassian-labs/aws-resources/pull/4) into aws-resources).

After the bucket is set up (in DataCenterFormula), the attachment/avatar data is then copied from the shared home into the bucket (in SharedHomeFormula).  I played around with this for a while to get something that copies the large datasets quickly.

It also configures Jira to look for avatars/attachments in the newly created S3 bucket (in DataCenterNodeFormula) via the new filestore-config.xml file.


src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/aws/AwsCli.kt
```diff
@@ -12,7 +12,14 @@ import java.util.concurrent.ConcurrentHashMap
  *
  * @since 2.15.0
  */
-class AwsCli {
+class AwsCli (val cliVersion: String = "2.9.12") {
```
MatthewCochrane, line 15: I've changed the default version to 2.9.12.  The bulk copying in `SharedHomeFormula.kt` wasn't reliable with the V1 CLI because it only supports legacy retry mode not [standard retry mode](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-retries.html#cli-usage-retries-modes-standard.title).  The new code can install a V1 or V2 cli and I've tested both, though it uses 2.9.12 by default everywhere now which I'm not sure if you'll be ok with or not.

src/main/resources/aws/object-storage.yaml
```diff
@@ -0,0 +1,42 @@
...
+    Type: String
+  AWSAccount:
+    Type: String
+Resources:
+  S3Storage:
+    Type: AWS::S3::Bucket
+    Properties:
+      BucketName:
+        Ref: S3StorageBucketName
+      LifecycleConfiguration:
```
MatthewCochrane, line 14: The bucket contents get automatically deleted by AWS in 1 day using a lifecycle configuration.  This reduces the complexity and increases the reliability of deleting the files in the bucket. The alternative of deleting files from the JPT code is both slow and error-prone (ie if JPT stops then the files would never get deleted).

MatthewCochrane: Another side effect of this is that we need permission to set the lifecycle configuration for buckets.  This will require adding a couple of permissions to the role JPT runs as.

lib/aws-resources-1.11.3-SNAPSHOT.jar
MatthewCochrane, line 1: I've hacked this in to run it, similar to the libs dir in JPT, but will remove and update the library version when the [aws-infrastructure PR](https://github.com/atlassian-labs/aws-resources/pull/4) is merged.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/home/SharedHomeFormula.kt
```diff
@@ -37,6 +47,31 @@ internal class SharedHomeFormula(
...
             ubuntu.install(it, listOf("nfs-kernel-server"))
             it.execute("sudo echo '$localSharedHome $localSubnet(rw,sync,no_subtree_check,no_root_squash)' | sudo tee -a /etc/exports")
+
+            if (s3StorageBucketName != null && storeInS3.isNotEmpty()) {
+                AwsCli(awsCliVersion).ensureAwsCli(it)
+                if (storeInS3.contains(ATTACHMENTS)) {
+                    // Copy the attachment data from NFS onto S3
+                    // Use xargs to split the work across 30 concurrent jobs.  This is much faster than a single job.
+                    it.safeExecute(
+                        cmd = "export AWS_RETRY_MODE=standard; export AWS_MAX_ATTEMPTS=10; cd $localSharedHome/data && ( find attachments -mindepth 1 -maxdepth 1 -type d -print0 | xargs -n1 -0 -P30 -I {} aws s3 cp --recursive --only-show-errors {}/ s3://$s3StorageBucketName/{}/ )",
```
pczuj, line 57: Not necessarly actionable in this PR - just FYI / future work note for other reviewers.

This is hard to unit test. I created suggestion for us to consider: https://ecosystem.atlassian.net/browse/JPERF-1122

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/JiraSharedStorageResource.kt
```diff
@@ -0,0 +1,8 @@
+package com.atlassian.performance.tools.awsinfrastructure.api.jira;
+
+/**
+ * Describes a set of resources used by Jira and stored in shared storage (eg. shared home or S3).
+ */
+enum class JiraSharedStorageResource {
```
pczuj, line 6: Can this enum change in the future? If so then `enum` is not a great option for it, as changes to it would break API.

Usually polymprphism is a good way to avoid `enum` usage. If you want I could help with figuring out how a polymprphic model could look like here.

We could proceed with `enum` if it's very unlikely for the value set to change.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/aws/AwsCli.kt
```diff
@@ -21,21 +28,47 @@ class AwsCli {
...
+        )
+        ssh.execute("unzip -n -q awscli-bundle.zip")
+        ssh.execute(
+            cmd = "sudo ./awscli-bundle/install -i /usr/local/aws -b /usr/local/bin/aws",
+            timeout = Duration.ofSeconds(60)
+        )
+    }
+
+    private fun installV2Cli(ssh: SshConnection) {
+        // Instructions from https://docs.aws.amazon.com/cli/latest/userguide/getting-started-version.html
```
pczuj, line 60: Nit: Wouldn't it be better as a method kdoc? It would be more clear that the comment is about the whole method and not only line 61

pczuj:
I think those changes look good. Before I approve I'd like to consult with the team if the the `enum` usage is acceptable, as it's currently my only concern. Hopefully we will be able to sync on that tomorrow.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/home/SharedHomeFormula.kt
MatthewCochrane: Yeah it's a bit dense that line.  We could likely achieve something similar if we could write jvm code to run on the remote as in your suggestion.  That would make it much more testable too.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/JiraSharedStorageResource.kt
MatthewCochrane: It's definitely possible for this to change.  Is there an example of the polymorphic approach you're suggesting and I could have a go at implementing it?

pczuj: What I was thinking about was a model where a polymorphic replacement for `JiraSharedStorageResource` could be an extension to `JiraHomeSource`. Both operations based on it that you implemented so far are done on Jira home, so I believe that it should logically fit there.

Each Jira resource type (avatars, attachments) could be implemented as wrapper which would delegate to any dataset. It would first let the delegate configure the home on remote and then it would create S3 bucket and upload selected  artifacts from Jira home there. Then it could append (or create) `filestore-config.xml`.

When I think about it now this may seem to be a bit of implementation changes. I'd personally still go for it, though maybe there is a simpler way to avoid `enum`s that the team can think of. Alternatively maybe `enum` are good enough for now - still need to consult with the team (we have a standup in 1h).

pczuj: I was trying to show the idea with this change set, however I wasn't able to finish it: https://github.com/atlassian/aws-infrastructure/pull/168

pczuj:
This won't be compatible with our method of saving datasets, as it assumes that jirahome from local file system is enough for it.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/home/SharedHomeFormula.kt
pczuj, line 57: Are we planning to share this upload method with customers? In JPT we like to have references to documentation in the code, as we want the automation to simulate the admin.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/JiraSharedStorageResource.kt
pczuj: @MatthewCochrane your alternative solutions to getting rid of `enum` usage that you posted in [this comment](https://github.com/atlassian/aws-infrastructure/pull/168#issuecomment-1566580555) should be fine. Please proceed with any solution that will work for you and don't be blocked by my suggestions. The main concern was about `enum` in API and usage of polymorphism is not mandatory.

MatthewCochrane:
Thanks for [this comment](https://github.com/atlassian/aws-infrastructure/pull/166#issuecomment-1564335575), thats a good point that it won't be compatible with the saving datasets code.  That actually looks quite involved to update.  Would it make sense to not allow dataset saving if attachments or avatars are stored in S3?

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/home/SharedHomeFormula.kt
MatthewCochrane: At the moment we will be recommending that customers use AWS DataSync instead of this approach.  If we do end up suggesting this approach to customers, then I'll be sure to update the code with a link to documentation.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/JiraSharedStorageResource.kt
MatthewCochrane: Thanks, have pushed a change.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/home/SharedHomeFormula.kt
pczuj: Would it be possible to use the AWS DataSync in our code? Our [approach](https://stash.atlassian.com/projects/JIRASERVER/repos/jira-performance-tests/browse/docs/APPROACH.md) was always to imitate admins, which has many benefits.

pczuj:
> would it make sense to not allow dataset saving if attachments or avatars are stored in S3?

If we can't simply make it work I think it would be good to at least somehow detect it and throw an exception. My biggest concern is that someone wouldn't be aware of saving an incomplete dataset.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/JiraSharedStorageResource.kt
pczuj, line 6: With the recent change this is no longer used. Should we remove it?

pczuj (changes requested):
- [x] Remove `JiraSharedStorageResource`
- [ ] Consider using AWS DataSync as we recommend to our admins
- [x] Don't allow to save dataset with missing attachments / avatars
- [x] Consider joining new configuration into one object
- [ ] Update changelog

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/JiraSharedStorageResource.kt
MatthewCochrane: Absolutely, thanks.

pczuj:
Also please add [CHANGELOG.md](https://github.com/atlassian/aws-infrastructure/blob/master/CHANGELOG.md) entry

MatthewCochrane:
@pczuj Is there a way to use the [release-1.12.0-alpha](https://github.com/atlassian-labs/aws-resources/releases/tag/release-1.12.0-alpha) release or should I wait for a non-alpha release of aws-resources so I can removed the hacked lib?

pczuj:
The release attempt already happened, but it failed: https://github.com/atlassian-labs/aws-resources/actions/runs/5088283216/jobs/9146117217

Needs investigation

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/home/SharedHomeFormula.kt
MatthewCochrane: I'm open to your thoughts here but I think using DataSync in JPT may be overkill.  DataSync is recommended to customers because it's very robust, and reliable for large instances (eg. 1TB+ of attachments).
To setup DataSync requires quite a lot of work and in general a customer with only a few GB (or 10's of GBs) of data like in the JPT datasets would use an approach like `aws cp`.  We actually expect some customers with smaller instances to use this approach (ie `aws cp`) instead of DataSync.  Perhaps this is more an issue with our documentation and we should look at recommending this approach there as well.

MatthewCochrane:
@pczuj  As we discussed in [your PR](https://github.com/atlassian/aws-infrastructure/pull/168#issuecomment-1567839557), I tried the copy operation again with `max_concurrent_requests` set to 300.  I did some tests:

| Method                                         | Throughput | CPU                              | Time to copy attachments (660k files, 6GB) |
| ---------------------------------------------- | ---------- | -------------------------------- | ------------------------------------------ |
| `aws cp` with `max_concurrent_requests = 300`  | ~20Mb/s    | 1 core maxed, ~2% in other cores | ~ 40 mins                                  |
| `aws cp` with `max_concurrent_requests = 3000` | ~20Mb/s    | 1 core maxed, ~2% in other cores | ~ 40 mins                                  |
| `aws cp` with `xargs`                          | ~200Mb/s   | High usage among all cores       | ~3.5 minutes                               |

I can only speculate as to the cause but it looks like even though `aws cp` offloads work to multiple threads, there's a main thread that is becoming a bottleneck. So when we start many copies of `aws cp` it can spread whatever work was the bottleneck out over multiple threads and get a higher throughput.  Here are some resources from AWS that led me to this approach.

- https://aws.amazon.com/blogs/apn/getting-the-most-out-of-the-amazon-s3-cli/ (especially example 1)
- https://repost.aws/knowledge-center/s3-improve-transfer-sync-command

pczuj:
aws-resources-1.12.0 is available: https://packages.atlassian.com/maven-central/com/atlassian/performance/tools/aws-resources/1.12.0/

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/home/SharedHomeFormula.kt
pczuj: We want our datasets in JPT to be of the scale of our customers.

If we want to recommend usage of `aws cp` to our customers then I think it should be fine for JPT as well. Let's just be consistent, because:
 * If we will have problems with the recommendation here, then our customers may also have those.
 * In case we will improve our recommendation we will also want to apply it in JPT to benefit.
 * With using different approach then recommendation we will face problems that our customers don't have.

Is there something problematic in DataSync that makes it hard to use in JPT specifically?

pczuj:
WRT `max_concurrent_requests` tests:

Sad to see the results. Usage of the flag would be much simpler. Anyway, if the process fork approach is faster, then we may want to use it. If this is some optimisation that is not part of the recommendation I'd suggest to documment it somehow in the code, so that we can reconsider it in the future without any doubt where did the `xarg` usage came from.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/home/SharedHomeFormula.kt
MatthewCochrane: I hear what you're saying around wanting to imitate Jira admins, it's a good form of added testing and understanding what our customers experience. At the moment, our documentation recommends DataSync, however it doesn't provide a definitive solution.  It leaves the decision of how to copy data to the admin and doesn't provide any specific commands.  Here's the relevant documentation:
image.

It's definitely possible to implement DataSync, though it will require more setup and configuration. In addition it may be prohibitively slow for the JPT use case (ie take closer to 40 minutes to copy default dataset). To set it up, we'll need to create a DataSync agent (on a new EC2 instance), configure source and destination locations, and set up task execution.

MatthewCochrane:
Hey @pczuj, I found some time to come back to this PR and update it.

I've rebased it on master and made a couple of small changes/fixes since the last time you looked at it.
- I [changed which role the bucket policy is granted to.](https://github.com/atlassian/aws-infrastructure/pull/166/commits/547049def01d8ca0da48c265b22e56e5d1c6675b) It was updated to the role of the EC2 instance profile the tests run as. This ensures the instance under test can access (and delete objects in) the bucket with avatars/attachments.
- I [increased the lifespan](https://github.com/atlassian/aws-infrastructure/pull/166/commits/48cd3dc2b17ac8b3559408bafb872febf41ee214) of the stack with the avatar/attachment S3 bucket to four days. This lets the bucket lifecycle policy delete the objects before the housekeeping plan tries to clean it up. AWS can take multiple days to perform the actual deletion, thus the extra buffer here.

In addition to these changes, to run, the `JPT-dev` role will also require the following permissions:
- s3:GetBucketPolicy
- s3:PutBucketPolicy
- s3:DeleteBucketPolicy
- s3:PutLifecycleConfiguration

Is there a process for adding these?

pczuj:
> Is there a process for adding these?

Yes, we store the permissions in https://github.com/atlassian-labs/aws-resources/blob/master/src/main/resources/iam-policy.json

MatthewCochrane:
> > Is there a process for adding these?
> 
> Yes, we store the permissions in https://github.com/atlassian-labs/aws-resources/blob/master/src/main/resources/iam-policy.json

Thanks, I added a PR for this [here](https://github.com/atlassian-labs/aws-resources/pull/16).

pczuj (approved):
Those changes look good to me. New API should be fine to maintain and all the internal logic is refactorable if needed.

Would be good to see @mgrzaslewicz and/or @dagguh review, as they will maintain JPT further while the rest of tech team will very likely move to other streams.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/jira/home/SharedHomeFormula.kt
mgrzaslewicz: This should be improved. Preference effort/quality starting with lowest
- extracted into a separated method
- injected as an object of a new class and unit tested
- @pczuj's suggestion https://ecosystem.atlassian.net/browse/JPERF-1122

CHANGELOG.md
```diff
@@ -23,6 +23,11 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-3.1.1...master
 
+### Added
+- Add `DataCenterFormula.Builder.jiraSharedStorageConfig` function to configure DC tests to store attachments/avatars in an S3 bucket.
+- Add `JiraSharedStorageConfig`.
+- Update AWSCli to support V2 and use V2.9.12 by default.
```
mgrzaslewicz, line 29: When 2 CLI versions dropped, this could become an implementation detail that could be skipped in changelog

mgrzaslewicz (changes requested):
There are a few improvements to be done. Happy to approve as soon as those are addressed

dagguh:
> Yes, we store the permissions in https://github.com/atlassian-labs/aws-resources/blob/master/src/main/resources/iam-policy.json

Nope, the `aws-resources` stores permissions needed by its APIs. If `aws-infrastructure` needs more permissions for its own APIs, it should have its own policy list here.

PR #168, opened 2023-05-26
Add draft of S3 avatars implemented as JiraHomeSource wrapper

pczuj:
This doesn't really work without `FilestoreConfigAppender` and `s3StorageProvider` implementations. It also wasn't really tested, however it's a demonstration of how the feature could be implemented using polymorphism on top of `JiraHomeSource`.

`FilestoreConfigAppender` interface is rather poor. Should probably be refined into something better modeled.

I used `aws s3 sync`, as it should better fit our use case where we actually want to ensure all files are present in S3 storage.
I used `StorageLocation` as it knows how the URI should look like and stores related AWS region.
Not sure if `aws configure set default.s3.max_concurrent_requests 300` can replace the usage of `xargs` to split the processes. I think it's worth testing, as this approach may be simpler.

Feel free to rebuild this code and reuse it's elements elsewhere.

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/AvatarsInS3.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/AvatarsInS3.kt
@@ -1,0 +1,59 @@
+package com.atlassian.performance.tools.awsinfrastructure.api.jira
+
+import com.atlassian.performance.tools.aws.api.StorageLocation
+import com.atlassian.performance.tools.awsinfrastructure.api.aws.AwsCli
+import com.atlassian.performance.tools.infrastructure.api.jira.JiraHomeSource
+import com.atlassian.performance.tools.ssh.api.SshConnection
+import java.time.Duration
+import java.util.function.Supplier
+
+class AvatarsInS3(
+    private val delegate: JiraHomeSource,
+    private val s3StorageProvider: Supplier<StorageLocation>,
+    private val filestoreConfigAppender: FilestoreConfigAppender
+) : JiraHomeSource {
+    private val associationTarget = "avatars"
+
+    override fun download(
+        ssh: SshConnection
+    ) = delegate
+        .download(ssh)
+        .also { jiraHomeLocation ->
+            val s3Bucket = s3StorageProvider.get()
+            uploadToS3(ssh, jiraHomeLocation, s3Bucket)
+            updateFilestoreConfig(ssh, jiraHomeLocation, s3Bucket)
+        }
+
+    private fun uploadToS3(
+        ssh: SshConnection,
+        jiraHomeLocation: String,
+        s3Bucket: StorageLocation
+    ) {
+        AwsCli("2.9.12").ensureAwsCli(ssh)
+        val avatarsLocation = "$jiraHomeLocation/data/avatars"
+        val s3AvatarsUri = s3Bucket.uri.resolve("avatars")
+        ssh.execute("aws configure set default.s3.max_concurrent_requests 300")
+        ssh.execute(
+            cmd = "AWS_RETRY_MODE=standard AWS_MAX_ATTEMPTS=10 aws s3 sync $avatarsLocation $s3AvatarsUri",
+            timeout = Duration.ofMinutes(10)
+        )
+    }
+
+    private fun updateFilestoreConfig(
+        ssh: SshConnection,
+        jiraHomeLocation: String,
+        s3Bucket: StorageLocation
+    ) {
+        val filestoreConfigLocation = "$jiraHomeLocation/filestore-config.xml"
+        filestoreConfigAppender.append(ssh, filestoreConfigLocation, s3Bucket, associationTarget)
+    }
+
+    interface FilestoreConfigAppender {
+        fun append(
+            ssh: SshConnection,
+            filestoreConfigLocation: String,
+            s3Bucket: StorageLocation,
+            target: String
+        )
+    }
+}


pczuj:
Context: In Hadrons team during today's standup @dagguh suggested to show the suggestion as a code.

I didn't manage to create complete code as I had some gaps in my knowledge about the feature:
 * What is "target" in the `<association>` of  `filestore-config.xml`? Is it always the same as the directory name under `<jirahome>/data`?
 * Can there be reasonably more than one `<filestores>` in  `filestore-config.xml`?
 * Is the `filestore-config.xml` used for other features?
 * Could we just upload everything in `<jirahome>/data` directory to S3?
 * What are the differences between avatars and attachments that are affecting the upload+config logic for them? From the code analysis I wasn't able to spot anything but the directory name.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/AvatarsInS3.kt
```diff
@@ -0,0 +1,59 @@
...
+    private val s3StorageProvider: Supplier<StorageLocation>,
+    private val filestoreConfigAppender: FilestoreConfigAppender
+) : JiraHomeSource {
+    private val associationTarget = "avatars"
+
+    override fun download(
+        ssh: SshConnection
+    ) = delegate
+        .download(ssh)
+        .also { jiraHomeLocation ->
```
MatthewCochrane, line 21: How would we ensure that uploadToS3 runs only on the shared home (NFS) and that updateFilestoreConfig runs only on application nodes?

MatthewCochrane:
Thanks for taking a look and writing this example!
Just so I understand, in the `DataCenterFormula` interface that I proposed, we pass an EnumSet of options into DataCenterFormula when it's constructed.  In this polymorphic approach does that mean we'd define combinations of `JiraHomeSource`'s? For example `AvatarsInS3`,  `AttachmentsInS3` and  `AvatarsAndAttachmentsInS3` and pass that into the `DataCenterFormula` constructor?

In this approach, how would `DataCenterFormula` would know whether or not to [create the new stack for the bucket](https://github.com/atlassian/aws-infrastructure/pull/166/files#diff-58e4eecaa5090c8d01e298bf8e96492d5c9ded10ee1ac8729975e17abfee4875R149-R163)?  At the moment it checks to see if the EnumSet is empty.


To answer your questions:

> What is "target" in the <association> of filestore-config.xml? Is it always the same as the directory name under <jirahome>/data?

Target is the the target of that association.  It can be either `avatars` or `attachments` and indicates to jira that it should look for that target on the file-store specified. https://hello.atlassian.net/wiki/spaces/SYNERGY/pages/2480715686/Simplified+Configuration+Format

> Can there be reasonably more than one <filestores> in filestore-config.xml?

Yep.  Though I don't see any need to configure a second one for the JPT tests at this point.

> Is the filestore-config.xml used for other features?

It's only used to tell Jira where to look for avatars/attachments at this point.  Could be extended if we store other data there in the future, though there are no projects lined up to do that at this point (nor have I heard discussion to do it).

> Could we just upload everything in <jirahome>/data directory to S3?

That is possible.  Not sure of the advantage of that though.


> What are the differences between avatars and attachments that are affecting the upload+config logic for them? From the code analysis I wasn't able to spot anything but the directory name.

They get configured differently in filestores-config.xml (they have different targets).  Also the uploading logic is different.  To get reasonable performance I had to split up avatars into different directories so I can run parallel jobs with XARGS.


Also note that I did try using `aws s3 sync` and `max_concurrent_requests` but it was still too slow. I was getting about ~1.6MiB/s and the attachments in ENTERPRISE_UNIMODAL are about 5.9GB so that would take about an hour to copy whereas the xargs copy took about 3.5 mins.  I didn't set the `max_concurrent_requests` quite as high though.


If the issue with the Enum is that it's in API and adding new items can break backward compatibility, just brainstorming, some other approaches could be:
- use two booleans (a bit ugly but adding new ones won't break API compatibility)
- limit the scope of this enum to just avatars/attachments (this could mean renaming it to be more specific).  If we did move other items to S3 we'd use a different parameter (note: although it's possible to see an extension to what is stored in S3, it is unlikely to happen for quite a long time (ie years) as there's be no discussion of it so far).

pczuj:
> Not sure of the advantage of (uploading everything in /data directory to S3) though.

The desired advantage would be simplicity of the implementation which would also affect maintainability of the code around the feature. In the future we'd like to refactor `DataCenterFormula` and having every component of it as simple as possible will definitely help.

> I didn't set the `max_concurrent_requests` quite as high though.

In [the original bash script](https://github.com/atlassian/aws-infrastructure/pull/166/files#diff-5c2dd8a96349c090ab1fa1ddb4b54bf82c029c8bd351534b1381acc0501cb8a7R57) we are using 30 xarg processes and the default `max_concurrent_requests` is 10. This is why I used 30 * 10 = 300.

Not sure if 300 is allowed, as I can't find any documented boundry for that property. I think it's worth checking if setting it to some high value wouldn't be good enough, because the complexity of the code is drastically reduced and we are relying on dedicated feature for the purpose, so we can easily rely on automated tests done by AWS cli devs.

> (...) other approaches could be
>  * use two booleans (a bit ugly but adding new ones won't break API compatibility)

I think that doesn't need to be ugly and it may be the simple solution that we need. We could have a single object which would hold the configuration of the feature including 2 booleans. Not very OOP approach, but it should be good enough for me.

MatthewCochrane:
> Not sure if 300 is allowed, as I can't find any documented boundry for that property. I think it's worth checking if setting it to some high value wouldn't be good enough, because the complexity of the code is drastically reduced and we are relying on dedicated feature for the purpose, so we can easily rely on automated tests done by AWS cli devs.

I'll give it a try using `aws sync` and `aws cp` with `max_concurrent_requests` set to 300 and see how it goes!

> I think that doesn't need to be ugly and it may be the simple solution that we need. We could have a single object which would hold the configuration of the feature including 2 booleans. Not very OOP approach, but it should be good enough for me.

Ok, sounds good!

MatthewCochrane:
> Yes, the usage would be to wrap JiraHomeSource passed to DataCenterFormula in our "S3 storing" wrapper. We could have different wrappers for each artifact (AvatarsInS3, AttachmentsInS3) as I suggested in example code I made, however we could also just have a single class for the whole feature, e.g. S3JiraHome. This single class could have the 2 booleans enabled in constructor/builder.
You can check [JiraUserPasswordOverridingDatabase](https://github.com/atlassian/infrastructure/blob/4eaf7eb8bd386db9841555a25fef47f6597bf57a/src/main/kotlin/com/atlassian/performance/tools/infrastructure/api/database/passwordoverride/JiraUserPasswordOverridingDatabase.kt) for similar usage pattern. We are using it [like this](https://stash.atlassian.com/projects/jiraserver/repos/jira-performance-tests/browse/jira-private-datasets/src/main/kotlin/com/atlassian/jira/test/performance/datasets/AdminPassword.kt?at=a1ab7c1306cc038325c8558268ea3d555a901051), e.g. [here](https://stash.atlassian.com/projects/jiraserver/repos/jira-performance-tests/browse/jira-software-tests/src/test/kotlin/com/atlassian/jira/test/performance/ManualIT.kt?at=a1ab7c1306cc038325c8558268ea3d555a901051#110)

Ok thanks for the explanation! That makes sense.  I see most of the pieces to make this work with the approach you outlined.  There are still a couple missing though.  The main one is probably [determining which node we're on](https://github.com/atlassian/aws-infrastructure/pull/168#discussion_r1208890062).  Since you mentioned the two-boolean approach looks good-enough I'll update to that for now.

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/AvatarsInS3.kt
pczuj: Could we not ensure that? I was thinking if `aws s3 sync` wouldn't resolve that problem for us. `sync` executed twice shouldn't duplicate any work.

PR #173, opened 2023-07-10
Issue/jperf 1184 stop docker containers for real

ewefie:
Add sudo to CustomDatasetSource.stopDockerContainers. 
We suspect that when we fail to close the MySQL container and proceed with archiving the database, the data inside might be corrupted.

### DIFF
--- a/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/CustomDatasetSource.kt
+++ b/src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/CustomDatasetSource.kt
@@ -94,7 +94,7 @@ class CustomDatasetSource private constructor(
     }
 
     private fun stopDockerContainers(host: SshHost) {
-        Ssh(host, connectivityPatience = 4).newConnection().use { it.safeExecute("docker stop \$(docker ps -aq)") }
+        Ssh(host, connectivityPatience = 4).newConnection().use { it.execute("sudo docker stop \$(sudo docker ps -aq)") }
     }
 
     override fun toString(): String {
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/CustomDatasetSource.kt
```diff
@@ -94,7 +94,7 @@ class CustomDatasetSource private constructor(
     }
 
     private fun stopDockerContainers(host: SshHost) {
-        Ssh(host, connectivityPatience = 4).newConnection().use { it.safeExecute("docker stop \$(docker ps -aq)") }
+        Ssh(host, connectivityPatience = 4).newConnection().use { it.execute("sudo docker stop \$(sudo docker ps -aq)") }
```
ewefie, line 97: I changed `safeExecute` to `execute` on purpose. We should be notified that something went wrong during this operation.

src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/CustomDatasetSourceIT.kt
```diff
@@ -0,0 +1,65 @@
...
+
+        //when
+        datasetSource.store(aws.customDatasetStorage(nonce).location)
+
+        //then
+        val databaseSsh = Ssh(datasetSource.database.host)
+        val use = databaseSsh.newConnection().use { ssh ->
+            return@use ssh.execute("sudo docker ps -q")
+        }
+        assertThat(use.output).isEmpty()
```
dagguh, line 35:
```suggestion
        val containerList = databaseSsh.newConnection().use { ssh ->
            ssh.execute("sudo docker ps -q")
        }
        assertThat(containerList.output).isEmpty()
```

src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/DatasetCatalogue.kt
```diff
@@ -77,6 +78,17 @@ class DatasetCatalogue {
             )
         }
 
+    fun smallJiraNine(): Dataset = DatasetCatalogue().custom(
```
dagguh, line 81: New API. Gotta changelog it

dagguh (changes requested):

dagguh (approved):

PR #175, opened 2023-09-07
JPERF-273 Switch to hooks preserving backwards compatibility

mgrzaslewicz:
Consume infra 3aacf87

Next steps:
* pristine Jira DC Plan unencumbered by `LegacyAwsInfrastructure` or CFN
* bridge from `JiraInstancePlan` to `JiraFormula` if possible
* bridge the other way around if possible
* can `SambaSharedHome` really work with just `TcpNode`?
* do we actually need `Reports` transport via S3?
* do we need bulk/directory transport for `Reports`?
* add jpt-workspace as CI artifact
* make hook collections append-only (`PreInstanceHooks` already are)


src/main/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/LegacyAwsInfrastructure.kt
```diff
@@ -0,0 +1,261 @@
...
+    private val databaseVolume: Volume,
+    private val provisioningTimout: Duration
+) : Networked, AutoCloseable {
+    private val logger: Logger = LogManager.getLogger(this::class.java)
+    private val nonce = UUID.randomUUID().toString()
+    private val sshKey: SshKey by lazy { provisionKey() }
+    private val provisionedNetwork: ProvisionedNetwork by lazy { networking.get() }
+    private val network: Network by lazy { provisionedNetwork.network }
+    private val provisioning: ProvisionedStack by lazy { provisionStack() }
+    private val deprovisioning: CompletableFuture<*> by lazy {
```
pczuj, line 55: Probably not important, but accessing this will first provision and then deprovision, so if we call `close` without `serveTcp` we would waste some time and resources - possibly without knowing.

PR #179, opened 2023-09-13
JPERF-1352 Upgrade jdbc connector

mgrzaslewicz:
Version 8.0.33 can be used both with mysql 5 and 8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ Dropping a requirement of a major version of a dependency is a new contract.
 ## [Unreleased]
 [Unreleased]: https://github.com/atlassian/aws-infrastructure/compare/release-3.1.1...master
 
+### Fixed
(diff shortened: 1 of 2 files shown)


pczuj:
The actual rationale behind this change is that Jira since 8.12 supports both MySQL connector versions (5.1 and 8.0) for MySQL 5.7, so we can support both MySQL 8.0 and 5.7 using this connector.

Source: https://confluence.atlassian.com/adminjiraserver0812/supported-platforms-1026528236.html

With this bump we are likely losing support for Jira 8.11 and below, however our datasets are generated on Jira 8.20.0 anyway, so we believe it's not worth keeping the compatibility. That is especially that Jira 8.11.0 is already EOL and even 8.20.0 will be EOL soon.

Source: https://confluence.atlassian.com/support/atlassian-support-end-of-life-policy-201851003.html

pczuj (approved):

PR #183, opened 2023-12-22
Use TaskScope

dagguh:
When bypassing the `TaskScope`, we ended up as having insufficient metadata, which ended up as "unknown" tasks in GUI:
image
image


src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/LogConfiguration.kt
```diff
@@ -30,11 +30,6 @@ internal class LogConfiguration(
                     null
                 )
             },
-            logToFile(
```
mgrzaslewicz, line 33: Why not have both (events and logs)?

dagguh: Unnecessary tax and dependency. If a consumer wants logs, they can subscribe and log. They don't need it by default thanks to GUI.

dagguh: Also, that behavior change is caused by switching from `TaskTimer` to `TaskScope`, not from this log config. This is just removing dead code here.

mgrzaslewicz (approved):

dagguh:
After:
image

PR #190, opened 2024-06-13
Bump Gradle and stuff

dagguh:
Diff might be big, but commits should make sense.

PR #192, opened 2024-06-17
Avoid flaky `OracleJDK` in tests

dagguh:


dagguh:
Wut, this got merged with red CI...

dagguh:
The CI was correctly red.

It showed `Wrong JVM version! You are running with .. but JIRA requires at least 1.8 to run.`.
Just like in https://github.com/teamatldocker/jira/issues/86

The [Jira 7.6 docs](https://confluence.atlassian.com/adminjiraserver076/supported-platforms-945110772.html) said it supported AdoptOpenJDK - Java 8 - but it's a straight up lie.
This is from Jira 7 `check-java.sh`:
```
"$_RUNJAVA" -version 2>&1 | grep "java version" | (
        IFS=. read ignore1 version ignore2
        if [ ! ${version:-0} -ge "$_EXPECTED_JAVA_VERSION" ]
        then
           echo "*************************************************************************************************************************************"
           echo "**********     Wrong JVM version! You are running with "$ignore1"."$version"."$ignore2" but JIRA requires at least 1.8 to run.      **********"
           echo "*************************************************************************************************************************************"
           exit 1
        fi
    )
```
The `grep "java version"` is the culprit.
`java -version` for Oracle matches:
```
java version "1.8.0_131"
Java(TM) SE Runtime Environment (build 1.8.0_131-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.131-b11, mixed mode)
```
But Adopt does not:
```
openjdk version "1.8.0_412"
OpenJDK Runtime Environment (build 1.8.0_412-8u412-ga-1~20.04.1-b08)
OpenJDK 64-Bit Server VM (build 25.412-b08, mixed mode)
```

The fixed `check-java.sh` came only in Jira 8.

Summing up, Jira 7 works only with Oracle JDK 1.8. If we want to switch away from the flaky Oracle mess, we need to drop Jira 7 support in JPT, a breaking change and a major SemVer. We can do it, FYI @mgrzaslewicz.

dagguh:
PS. Fixed `master` protection to look at `build` job and not `build-check` (unnecessary and gone in https://github.com/atlassian/aws-infrastructure/pull/190).

PR #193, opened 2024-06-19
Stop testing Jiras which require Oracle JDK 1.8

dagguh:
`OracleJDK` is flaky as hell. We just cannot make the tests pass, so we can't even release 3.2.0 with it.

JSW supports alternative JDKs since 8.0.0..
No, [7.6] does not support it, see the [explanation]. JSM supports it since 3.16.0.

[7.6]: https://confluence.atlassian.com/adminjiraserver076/supported-platforms-945110772.html [explanation]


src/test/kotlin/com/atlassian/performance/tools/awsinfrastructure/api/jira/DataCenterFormulaIT.kt
```diff
@@ -47,174 +38,77 @@ class DataCenterFormulaIT {
...
+            .waitForRunning(true)
+            .configs(nodeConfigs)
+            .build()
+
+        val provisionedJira = dcFormula.provision(
+            investment = Investment(
+                useCase = "Test Data Center provisioning",
+                lifespan = lifespan
             ),
-            DataCenterJmxProvisioningTest(
```
dagguh, line 61: The `shouldProvisionDataCenter` used to spawn 2 threads, one for testing LB, one for testing JMX. I rolled it up into a single test.

dagguh:
This covers the mistake from https://github.com/atlassian/aws-infrastructure/pull/192
And I don't see any other way to release 3.2.0.

PR #196, opened 2024-12-03
DCPERF-747 Update the aws-resources dependency

jacek-foremski:
Update the dependency to allow overriding the resource_owner label in Investment


dagguh:
No need to release a dependency bump. The consumers of aws-infra can bump aws-resources on their own if they need the new API. They don't have to rely on transitive deps.

jacek-foremski:
Actually what I'm trying to do is to fix the binary incompatibility that I encountered when bumping the library in the consumer.
But that could potentially be fixed with https://github.com/atlassian-labs/aws-resources/pull/28. 
Would appreciate the review as I'm not that familiar with Kotlin.

dagguh:
Yeah, that's an unfortunate trap of Kotlin data classes. We started avoiding them in API. BTW possibly Java record classes would be a better fit. 
Releasing the compatibility fix in aws-infra:1.18.1 is enough. Only the consumers who bumped to the bugged 1.18.0 need to bump again. So this PR is not necessary.

